Issue #4, opened 2020-10-07
Evaluate well-known errors for compatibility posture


paddycarver:
Our well-known errors, like `ErrUnsupportedDataSource`, are currently set as error values, not types. We may want to evaluate making them less well-known, or returning `fmt.Errorf` wrapped versions, to force people to use `errors.As` instead of doing equality comparisons. This would give us a more flexible posture for future changes than we currently have.

bflad:
Apparently `ErrUnsupportedDataSource` disappeared at some point, but while there are other errors that could use this treatment, I'm not sure there's necessarily been a strong standalone need to introduce breaking changes for this since the API has been fairly stable. I think the best thing to do with this for now is to close this but always keep this consideration in mind for future changes. 👍

PR #42, opened 2020-11-10
Add the ability to unmarshal RawState.

paddycarver:
The tfprotov5.RawState type exposes state through a []byte containing
JSON when upgrading resource state with the UpgradeResourceState RPC. We
had not surfaced any JSON parsing or decoding in the exported API, so
providers would need to parse the JSON themselves.

To solve this, we use the same Unmarshal interface we use for parsing
JSON from DynamicValues, letting users receive a tftypes.Value from a
RawState without needing to worry about encoding... mostly.

There is one hitch. The Flatmap syntax doesn't have a canonical
translation (yet? As far as I know) to Terraform's type system. So we
sidestep that by returning an error, for the moment. In the future, we
could conceivably institute a mapping, though I'm not convinced it's a
good idea to, given the number of assumptions that would need to be
made. It is a larger undertaking to do that in a predictable, reliable
way. Provider developers should provide their own mapping of Flatmap to
tftypes.Value.

The Flatmap property should only ever be populated in one specific
scenario, as best I can tell:

1. The user _was_ using a version of Terraform below 0.12, and that is
the last version of Terraform to write the state file.
2. The user upgrades to 0.12+ and obtains a new release of the provider
3. The new release of the provider is built on plugin-go

At that point, the plugin-go-based provider would be the first version
of the provider to write the state after 0.12, meaning it would be its
responsibility to upgrade the state between the flatmap syntax and
Terraform's type system.

It is unknown how likely that scenario is, but I want to document it
here to avoid any confusion.

### DIFF
--- a/tfprotov5/state.go
+++ b/tfprotov5/state.go
@@ -1,5 +1,25 @@
 package tfprotov5
 
+import (
+	"errors"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5/tftypes"
+)
+
+// ErrUnknownRawStateType is returned when a RawState has no Flatmap or JSON
+// bytes set. This should never be returned during the normal operation of a
+// provider, and indicates one of the following:
+//
+// 1. terraform-plugin-go is out of sync with the protocol and should be
+// updated.
+//
+// 2. terrafrom-plugin-go has a bug.
+//
+// 3. The `RawState` was generated or modified by something other than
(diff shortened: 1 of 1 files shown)


alexsomesan:
Excellent !
Excellent

paultyng (approved):
LGTM, although not sure if a well known error is useful for just one case here and not the other.

paddycarver:
We can address that as part of #4, when we determine what our well-known error philosophy for this module is.

PR #46, opened 2020-11-12
Support more types in NewValue

paultyng:


tfprotov5/tftypes/value_test.go
```diff
@@ -58,19 +88,130 @@ func TestValueAs(t *testing.T) {
...
+			in:       NewValue(Number, (*uint)(nil)),
+			as:       numberPointerPointer(big.NewFloat(123)),
+			expected: numberPointerPointer(nil),
+		},
 		"number-null": {
-			in:       NewValue(Number, nil),
-			as:       big.NewFloat(123),
+			in: NewValue(Number, nil),
+			as: big.NewFloat(123),
+			// Shouldn't this be nil? unsure why we can't set that value?
```
paultyng, line 214: @paddycarver left this question here, curious why you shouldn't expect a nil here for the big float?

paultyng: We can drop this comment if its intentional, just seemed weird to me, not what I would have expected if you have a nil-able type.

```suggestion
```

paddycarver: I wanted the behavior to be consistent--one pointer, zero value, double pointer, nil. `big.Float` is weird because it's _normally_ passed around boy pointer, but I felt it was more consistent.

paddycarver (approved):
Let's do it, we can refactor and update documentation later if we want to, but the code is correct and the behavior is desirable.

EDIT: once you gofmt

PR #47, opened 2020-11-14
Also support floats in NewValue

paultyng:


paultyng:
I think I added non pointer types to the godoc, but I realize now I haven't tested those, not sure they actually work, probably do not as they aren't coded as such so that needs to be addressed. Perhaps in a follow-up PR?

paddycarver (approved):
Fair enough.

PR #50, opened 2020-12-08
Add support to Value.As for interface types

paultyng:
This will enable usage of things like mapstructure for easy bulk assignment of ttftypes values to structs.

### DIFF
--- a/tfprotov5/tftypes/value.go
+++ b/tfprotov5/tftypes/value.go
@@ -287,7 +287,7 @@ func NewValue(t Type, val interface{}) Value {
 //
 // For Lists, Sets, and Tuples, `dst` must be a pointer to a []Value or a
 // pointer to a pointer to a []Value. If it's a pointer to a pointer to a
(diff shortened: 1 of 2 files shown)


paddycarver (changes requested):
I think after some discussion, we want this to be a helper outside plugin-go. I'll leave this open until we have one available, but this really only makes sense in a handful of situations when certain conditions are met, and I'm worried people will reach for it in unsafe situations or when those conditions aren't met if we include it as part of the main module.

Really, converting complex or aggregate `Value`s is only safe when all the sub-`Value`s are known. I'm worried that because we've conditioned provider developers to expect a `map[string]interface{}`, they're going to reach for it and run into problems with unknown values at runtime, as interpolation (one of the major causes of unknown values) doesn't often get tested thoroughly.

paddycarver:
I'm going to close this out, with the recommendation to use something like [`asgotypes`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go-contrib/asgotypes) instead.

PR #58, opened 2021-02-04
Expose Types of tftypes.Values.

paddycarver:
Add a method to expose the tftypes.Type associated with a tftypes.Value.
This is necessary because tftypes.Value.Is is not always able to
reconstruct a type from comparison alone; for example, reconstructing
what types a map, object, list, set, or tuple uses when its
tftypes.Value is the unknown value is impossible.

But we can make it so people can't directly compare them with ==, which
we don't want them doing (we want them to use Is instead) by embedding a
`_ []struct{}` in each type, which cannot be compared and prevents the
types from being compared, raising a compile time error.

This does mean our primitive types are now structs, and therefore cannot
be constants, which is sad. In theory, someone could overwrite them and
change the output, but the amount of shenanigans one can get up to with
this is pretty limited. You could rename a type to another type:

```go
tftypes.DynamicPseudoType = tftypes.String
```

but you can't construct a new type out of whole cloth, because the
`primitive` type isn't exported and so you can't create new values for
it. So people could wreak some havoc here if they were malicious, but
this doesn't _seem_ to expose us to any new vectors of user error.

### DIFF
--- a/tfprotov5/tftypes/list.go
+++ b/tfprotov5/tftypes/list.go
@@ -6,6 +6,7 @@ import "fmt"
 // of the same type.
 type List struct {
 	ElementType Type
+	_           []struct{}
 }
 
 // Is returns whether `t` is a List type or not. If `t` is an instance of the
--- a/tfprotov5/tftypes/list.go
+++ b/tfprotov5/tftypes/list.go
@@ -6,7 +6,11 @@ import "fmt"
 // of the same type.
 type List struct {
 	ElementType Type
-	_           []struct{}
+
+	// used to make this type uncomparable
+	// see https://golang.org/ref/spec#Comparison_operators
+	// this enforces the use of Is, instead
+	_ []struct{}
 }
 
 // Is returns whether `t` is a List type or not. If `t` is an instance of the
--- a/.changelog/58.txt
+++ b/.changelog/58.txt
@@ -1,0 +1,11 @@
+```release-note:feature
+`tftypes.Value`s now have a `Type` method, exposing their `tftypes.Type`.
+```
+
+```release-note:breaking-change
+`tftypes.Value`s no longer have an `Is` method. Use `tftypes.Value.Type().Is` instead.
+```
+
+```release-note:breaking-change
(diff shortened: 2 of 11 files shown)


alexsomesan (commented):
This looks great so far!
I'm about to go.mod override this branch into the Kubernetes alpha provider and see how it fits in.
I'll report back in few hours.

kmoe:
Why do we want to prevent type comparison with `==`?

paddycarver:
Forcing comparisons through `Is` leaves us a more flexible API surface than `==`, because we have a more control over what is and is not equal then.

alexsomesan:
This works as expected for me! I wired this change in the Kubernetes alpha provider and was able to do an plan / apply / destroy cycle successfully. Works great for me! :+1:  Thanks!

paultyng:
@kmoe the specific example I think in this case is comparison to a tuple of any types, and a tuple of specific types, both of those could be asserted with `.Is` but not `==`.

tfprotov5/tftypes/list.go
```diff
@@ -6,6 +6,7 @@ import "fmt"
 // of the same type.
 type List struct {
 	ElementType Type
+	_           []struct{}
```
paultyng, line 9: can you please comment why these are added some someone doesn't inadvertently remove them.

this also prevents usage in map keys i believe as well, correct?

alexsomesan:
Changes in https://github.com/hashicorp/terraform-plugin-go/pull/58 will require updates to this PR if 58 merges first or the other way around, because walk and diff both check types via `Value.Is(...)` which was removed in #58.

tfprotov5/tftypes/list.go
paddycarver: Yes, Types can no longer be map keys after this change.

paddycarver: Nor can they be used in switch statements, or anywhere else that relies on some concept of comparison or uniqueness.

kmoe (approved):

Issue #56, opened 2021-01-26
Implement comparison of tftypes.Value in the sense of go-cmp


alexsomesan:
I've repeatedly run into an issue while writing unit tests for code that makes use of `tftypes.Number` values.

It turns out `big.Float` values which are backing the `tftypes.Number` values are pretty tricky to compare correctly. Depending on how the value has been constructed, certain private attributes in `big.Float` (like precision) end up being set to different values and fail the comparison of otherwise apparently equal values.

 Bellow is a quick demonstration of that aspect:
```go
package main

import (
	"fmt"
	"github.com/google/go-cmp/cmp"
	"math/big"
	"reflect"
	"strconv"
)

func main() {
	f, _ := strconv.ParseFloat("12.4", 64)
	a := new(big.Float).SetFloat64(f)

	b := big.NewFloat(12.4)

	c, _ := new(big.Float).SetString("12.4")

	x := cmp.Equal(a, b, cmp.Exporter(func(t reflect.Type) bool { return true }))
	y := cmp.Equal(c, b, cmp.Exporter(func(t reflect.Type) bool { return true }))
	z := cmp.Equal(a, c, cmp.Exporter(func(t reflect.Type) bool { return true }))

	fmt.Printf("a == b ? %v\n", x)
	fmt.Printf("c == b ? %v\n", y)
	fmt.Printf("a == c ? %v\n", z)
}
```

go-cmp allows for compared types to implement their own comparison mechanism, as described here:
https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal

It would be a nice UX improvement for `tftypes.Values` to implement the `Equal(..)` method so they would just work as expected out of the box with go-cmp.

Issue #53, opened 2021-01-04
Add Walk and Transform to tftypes


paultyng:
### Use cases

Walk:

* `IsFullyKnown` is a special case of `Walk` as an example
* Validation

Transform:

* Dynamic returns things that may possibly be lists as tuples, especially if there is a field mismatch due to limited info in the dynamic, and in some providers they need to be simplified to lists (kubernetes-alpha for example)
* Applying defaults to an object from schema

### Proposal

(similar to cty)

Since these implementations are not opinionated and leverage existing type/path info in the tftypes package, hopefully something that makes more sense in this repo as opposed to an outside one.

paddycarver:
For reference, I believe this is referring to [Walk](https://pkg.go.dev/github.com/zclconf/go-cty/cty/#Walk) and [Transform](https://pkg.go.dev/github.com/zclconf/go-cty/cty/#Transform)

paddycarver:
I can see an argument for including them in `tftypes`, I think more than anything I'd just want to think about stability. How likely are we to get the implementation right before v1? How likely is that implementation to change in the future?

I don't know the answer to those offhand, and they deserve some thought, but I'm not knee-jerk opposed to it.

paultyng:
Yeah no rush on these I think. Worth checking how much they changed in cty as well but the signatures look pretty straightforward.

paddycarver:
I think my concern is just that we'll want something that makes sense in the context of the SDK but not in the context of cty. For example, if we end up doing telemetry/logging/whatever, the function signature as it exists makes it awkward to inject those into the function being called, meaning its operations and decisions won't be observable. Which may not be the end of the world, but I think we should just think a bit about that kind of difference and whether there's anything else like that that may crop up.

paultyng:
In most scenarios I've used walk-like funcs (ie filepath, etc.) or things with callbacks you inject that stuff via closure, so it wouldn't be part of the signature, but happy to let this happen organically somewhat. We can see what @alexsomesan comes up with as well.

paddycarver:
Yeah, I think that's a really good call, and we could always have a helper to inject stuff into them, I suppose. Mostly just want to think through some of the future design directions a bit and see if the signatures and design all hold up.

PR #60, opened 2021-02-07
Walk, Transform, Diff, Equal, and String.

paddycarver:
This turned into a massive commit. I'm really sorry.

I set out to implement Walk and Transform. Walk was simple enough.
Transform was implemented quickly and easily enough. And then I tried to
test it.

Testing Transform reasonably necessitated all the other changes you see
here:

* We needed to be able to Diff Values to be able to express changes
  between Values so we could test that Transform made the changes we
  wanted it to.
* We needed to be able to represent AttributePaths as strings so we
  could have specific, useful test output. This also entailed making
  Values strings (AttributePaths can use Values as a step) and making
  Types have more specific strings (aggregate types now expose the
  subtypes that make up their type signature).
* In doing so, we kept accidentally overwriting our AttributePaths,
  meaning it was time for them to stop being modified in place and start
  being passed by value, with the steps defensively copied to avoid
  accidental modification.
* We needed the ability to detect whether Diffs were equal to know
  whether the changes we expected were the changes we actually got or
  not.
* We needed to make Values an AttributePathStepper so we could implement
  Diff, which is done by Walking one Value, and then retrieving the same
  sub-Value from the Value we're Diffing against to check if they're the
  same. We already have WalkAttributePath to retrieve part of a Go type
  based on the AttributePath passed in, we just needed to tell it how to
  index into a Value type.
* We needed the ability to Copy a Value type so that it was not sharing
  the same memory space anymore so our transformations didn't leak
  across tests.

The result is this PR of alarming size. The good news is a lot of these
new lines are tests.

### DIFF
--- a/tfprotov5/tftypes/attribute_path.go
+++ b/tfprotov5/tftypes/attribute_path.go
@@ -3,6 +3,8 @@ package tftypes
 import (
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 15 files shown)


paddycarver:
A lot of this code needs GoDoc strings. More tests are also needed, as are changelog entries.

These things are all forthcoming before we merge this, I just wanted to get the working code up so @alexsomesan could see and play with it.

alexsomesan:
Wow! Is it Christmas all over again?! This PR is just full of juicy goodies! 
I especially appreciate the changes to AttributePath to return values and not mutate itself. I really wish there .Errorf would expose the path somehow too. Although I guess I could just use it's String() method now to get to it.

Regardless, this is awesome. Thanks!
I will test these tomorrow and report impressions :)

paddycarver:
I am anticipating a separate PR once this is merged to make AttributePathError a bit more useful. Just trying to cut scope on this PR when I can 😂

alexsomesan:
Changes in #58 will require updates to this PR if 58 merges first or the other way around, because walk and diff both check types via Value.Is(...) which was removed in #58.

paddycarver:
Yup! Scattered some TODOs around this to take 58 into account. :)

paddycarver:
Fixes #53.

paddycarver:
Rebased after #58 and updated to use its Type() goodies.

tfprotov5/tftypes/attribute_path.go
```diff
@@ -26,6 +28,79 @@ type AttributePath struct {
...
+				return false
+			}
+		case ElementKeyInt:
+			oVal, ok := oStep.(ElementKeyInt)
+			if !ok {
+				return false
+			}
+			if oVal != aVal {
+				return false
+			}
```
paultyng, line 84: I think you could probably just collapse these down in to something like the below? The casting seems unnecessary and already handled by the equals? If value's aren't comparable though makes sense to pull that case out.

```go
default:
  if aStep != oStep {
    return false
  }
```

paultyng (approved):
Minor comment, but LGTM (once testing coverage added)

kmoe (approved):

tfprotov5/tftypes/walk.go
```diff
@@ -0,0 +1,209 @@
...
+func walk(path AttributePath, val Value, cb func(AttributePath, Value) (bool, error)) error {
+	shouldContinue, err := cb(path, val)
+	if err != nil {
+		var ape attributePathError
+		if !errors.As(err, &ape) {
+			err = path.NewError(err)
+		}
+		return err
+	}
+	if !shouldContinue {
```
alexsomesan, line 54: Thought experiment: I wonder if this should somehow surface that the visitation was interrupted before a complete exploration of all nodes could be finished. A special type of error maybe?

alexsomesan (approved):
Looks great!

I've actually made use of  Transform in the Kubernetes alpha provider and it works as expected. 

I did try to use Diff in a few places but it errors out with `incompatible types` when encountering dissimilar tuples. Because TF core always parses lists as tuples, it means any two lists with distinct amount of elements will make Diff err out. I don't think that's an issue with Diff itself, but rather an unfortunate combination of factors in my use-case.

tfprotov5/tftypes/walk.go
paddycarver: This seems to mirror the Go standard library walk behaviors, in the expectation that the original caller has some control over the callback? And given the callback is the one returning `shouldContinue`, they should know some things may be skipped? Similarly, the callback can return its own error that the caller can handle. I guess I'm not clear on when this would be used, or why.

paddycarver:
OK, finally finished adding all the tests on this. I think we're ready for final review and merge. Apologies for the monster diff. :( This kinda spiraled out of control.

paddycarver:
Fixes #53.
Fixes #56.

kmoe (approved):
Excellent! LGTM

PR #69, opened 2021-03-25
Fix panic in IsFullyKnown.

paddycarver:
Calling Value.IsFullyKnown would panic if the type of the value was an
aggregate type (List, Set, Tuple, Map, or Object) and if that value
itself was null.

This adds tests to reproduce those panics and a minor change to
IsFullyKnown that prevents the panics.

### DIFF
--- a/tfprotov5/tftypes/value.go
+++ b/tfprotov5/tftypes/value.go
@@ -609,6 +609,9 @@ func (val Value) IsFullyKnown() bool {
 	case primitive:
 		return true
 	case List, Set, Tuple:
+		if val.value == nil {
+			return true
+		}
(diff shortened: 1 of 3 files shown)


tfprotov5/tftypes/value.go
```diff
@@ -609,13 +609,19 @@ func (val Value) IsFullyKnown() bool {
 	case primitive:
 		return true
 	case List, Set, Tuple:
+		if val.value == nil {
```
alexsomesan, line 612: would it not be sufficient to do this check once right before `switch {` ?

paddycarver: Good call.

kmoe (approved):

PR #71, opened 2021-04-01
Add support for plugin protocol v6

mildwonkey:
This PR adds a new package, `tfprotov6`. It is *mostly* copy-pasted from `tfprotov5`. The interesting changes (besides just swapping 5 for 6 in imports everywhere) is the new `tfprotov6/internal/tfplugin6/tfplugin6.proto` file, taken directly from terraform's main branch, an assortment of renamed functions (`ValidateResourceTypeConfig` is now `ValidateResourceConfig`, etc - all the changes we discussed internally), and finally the interesting one, the new `NestedType` field inside an `Attribute` (with associated conversions in toproto and fromproto). 

If you want to see any of this in action, I have a very tiny implementation of the random provider with this branch vendored: https://github.com/mildwonkey/terraform-provider-random

I also have a fork of ptyng's sql provider using this; it works as a provider just fine but I'm still struggling with the acceptance tests, which required changes to the sdkv2. I'll keep poking at that; it'd be nice to show off a provider using these new features with working tests.


paddycarver (approved):
I think, as future work, we're going to want to refactor the JSON/MsgPack encoding out of the protocol-versioned packages and into tftypes, so we don't need to keep implementing them for each new protocol version, but I don't think it's reasonable to lump that in with this PR.

LGTM. 👍 Thanks!

PR #74, opened 2021-04-13
Implement OptionalAttributes for Objects.

### LINKED ISSUE #48: Add support for OptionalAttrs inside tftypes.Objects
### SDK version
build from main

### Use cases

Terraform (via go-cty) includes support for optional attributes inside object-type attributes. Extending terraform-plugin-go allows provider developers to write object-type `Attributes` (in place of blocks) that have optional attributes.

### Attempted solutions

The current method for handling situations like this is to create a `Block` with optional attributes, but there are times where an `Attribute` may be preferable to a `Block` (for example if explicitly setting a single object to null could be meaningful).

### Proposal

Something along the lines of this patch, but with better documentation and (I'd guess) also updating the `Is` function to compare `OptionalAttrs`:

```
diff --git a/tfprotov5/tftypes/object.go b/tfprotov5/tftypes/object.go
index 7b18c9e..fa8816b 100644
--- a/tfprotov5/tftypes/object.go
+++ b/tfprotov5/tftypes/object.go
@@ -9,6 +9,7 @@ import "encoding/json"
 // attribute names or types are considered to be distinct types.
 type Object struct {
        AttributeTypes map[string]Type
+       OptionalAttrs  []string
 }

 // Is returns whether `t` is an Object type or not. If `t` is an instance of
@@ -52,5 +53,15 @@ func (o Object) MarshalJSON() ([]byte, error) {
        if err != nil {
                return nil, err
        }
+
+       optionalAttrs, err := json.Marshal(o.OptionalAttrs)
+       if err != nil {
+               return nil, err
+       }
+
+       if len(optionalAttrs) > 0 {
+               return []byte(`["object",` + string(attrs) + "," + string(optionalAttrs) + `]`), nil
+       }
+
        return []byte(`["object",` + string(attrs) + `]`), nil
 }
```

I can open a PR if this is a direction you're interested in.

paddycarver:
This is another massive PR, sorry. I tried to separate out some of the
changes, but the code kept overlapping, unfortunately.

All our Types got an Equal method, so they can be compared using go-cmp.
This was necessary to make the tests pass. As part of this, the Type
interface was refactored to include an equals() private method, which
powers the Equal and Is methods on each type.

I also fixed some periods in error messages to make golint happy.

Of course, with these new methods, there was enough logic to add tests
to each Type, so I added some tests that exercised the equality checks.

I also added, which was the main thing I was trying to do here, an
OptionalAttributes property to Objects. These are attributes that _can_
be set on an object, but are not _required_ to be.

Because OptionalAttributes made the NewValue logic more complicated for
Objects, I wrote some tests for creating Objects with NewValue to test
the logic.

Finally, the new Object comparison logic was panicking during a
tfprotov5 / tfprotov6 test, because the Object had a nil AttributeTypes
map instead of an empty AttributeTypes map. I fixed it to use an empty
AttributeTypes map. In theory, I don't think we should ever see a nil
AttributeTypes map.

Fixes #48.

### DIFF
--- a/.changelog/74.txt
+++ b/.changelog/74.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+Added support for OptionalAttributes to tftypes.Objects, allowing for objects with attributes that can be set or can be omitted. See https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes for more information on optional attributes in objects.
+```
--- a/.changelog/74.txt
+++ b/.changelog/74.txt
@@ -1,3 +1,7 @@
 ```release-note:enhancement
 Added support for OptionalAttributes to tftypes.Objects, allowing for objects with attributes that can be set or can be omitted. See https://www.terraform.io/docs/language/expressions/type-constraints.html#experimental-optional-object-type-attributes for more information on optional attributes in objects.
 ```
+
+```release-note:enhancement
+Added Equal method to tftypes.Type implementations, allowing them to be compared using github.com/google/go-cmp.
+```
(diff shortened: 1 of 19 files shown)


paddycarver:
Tests are fixed by #72, will rebase this on top of it once it's merged.

kmoe (approved):

PR #76, opened 2021-04-15
[wip] Support nested dynamics in NewValue

paultyng:
This has the breaking test to illustrate the need for a better (recursive) dynamic allowance, not just at top level.

PR #79, opened 2021-04-18
Fix NewValue checks to be properly recursive

paddycarver:
Fix the checks that NewValue runs to ensure that a value can be used
with a Type to properly handle DynamicPseudoTypes nested in complex
types. Builds on and supersedes #76.

Should we make this part of the behavior for `.Is`? My gut says "no",
because `Is` is supposed to be about checking whether a type is
semantically similar, not that it can be used as another type. Maybe we
want a `Fulfills` method, similar to Is, that does this? I don't know
that this behavior needs to be exported, though...

Fixes a bug in #74 that would panic for complex types containing
DynamicPseudoTypes.

### DIFF
--- a/tftypes/list.go
+++ b/tftypes/list.go
@@ -67,6 +67,7 @@ func valueFromList(typ Type, in interface{}) (Value, error) {
 	switch value := in.(type) {
 	case []Value:
 		for pos, v := range value {
+			//TODO: this needs to be a recursive check to allow for nested dynamic
 			if !v.Type().Is(typ) && !typ.Is(DynamicPseudoType) {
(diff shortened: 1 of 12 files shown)


kmoe (approved):
Looks good. We can always export `UseTypeAs`/`Fulfils` or change `Is` behaviour later.

paddycarver:
I would like to, in 0.4.0, do some work on Is to make this cleaner and have less magic behavior:

* Is is a shallow check, that root types are the same type
* Equals is a deep check, that root types and all elements and attributes are the same type
* ConformsTo is a deep check, that root types and all elements and attributes conform (are equivalent, or a concrete type is being substituted for DynamicPseudoType)

I just don't want to hold up 0.3.0 for that work, it's a big enough release already. I'll write up an issue for it.

PR #86, opened 2021-06-11
Don't panic when String() is called on nil Value{}

kmoe:
While a `Value` without a `typ` will not be returned from `NewValue` except in case of error, `Value{}` is likely to appear in consumer code from time to time, e.g. in tests, as it is the nil value of the struct. 

This PR defines an arbitrary string representation for `Value{}`. The only non-trivial question here is what it should be.

Backwards-compatible because this was previously undefined behaviour.

### Bonus

This PR also adds lots of tests for `Value.String()`, which had little test coverage.

### DIFF
--- a/tftypes/value.go
+++ b/tftypes/value.go
@@ -46,6 +46,10 @@ type Value struct {
 func (val Value) String() string {
 	typ := val.Type()
 
+	if typ == nil {
+		return "tftypes.Value<>"
+	}
+
 	// null and unknown values we use static strings for
 	if val.IsNull() {
 		return typ.String() + "<null>"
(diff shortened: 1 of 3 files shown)


tftypes/value.go
```diff
@@ -46,6 +46,10 @@ type Value struct {
 func (val Value) String() string {
 	typ := val.Type()
 
+	if typ == nil {
+		return "tftypes.Value<>"
```
paddycarver, line 50: What if it we make it clearer that this is an _invalid_ Value, something that wasn't constructed as it was supposed to be?
```suggestion
		return "<invalid typeless tftypes.Value>"
```

kmoe: Good idea

paddycarver (approved):
Mind just adding a changelog entry in `.changelog/86.txt`?

PR #87, opened 2021-06-21
Better formatting for root AttributePaths.

paddycarver:
For empty AttributePaths (paths pointing to the root of the
type/whatever) don't prefix an empty colon, prefix it with `<root>:`.

### DIFF
--- a/.changelog/87.txt
+++ b/.changelog/87.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fixed AttributePaths pointing to the root of the value to be prefixed with `<root>:` instead of just with `:`.
+```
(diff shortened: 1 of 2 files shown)


kmoe (changes requested):
This modifies the string representation of an `AttributePath` only where it appears in an `AttributePathError`, so the value in the error message is inconsistent with `AttributePath.String()`. Prefer making it consistent by modifying String() instead. I've added unit tests for this in https://github.com/hashicorp/terraform-plugin-go/pull/89 so we have a spec for this behaviour.

paddycarver:
How is that latest push? It just omits the empty attribute path if there's an attribute path error.

kmoe (approved):

PR #88, opened 2021-06-21
Fix panic for Value.As(**string) and other types

remilapeyre:
The same bug that caused panics for

```go
var foo *big.Float
err := value.As(&foo)
```

affects also all the other types. This patch creates the underlying
value if we were given like it was done in 1df2217 for **big.Float.


paddycarver (approved):
LGTM, thanks! 👍

PR #90, opened 2021-06-24
Fix panic in tftypes.Value.Equal.

paddycarver:
If a non-empty tftypes.Value was compared to the empty value of a
tftypes.Value, it would cause a panic. Add tests to show the issue and
fix it.


bflad (approved):
Looks good to me 🚀

PR #94, opened 2021-08-09
tftypes: Enhance Type Comparisons Functionality

### LINKED ISSUE #80: Make type comparisons clearer
We currently have a _very_ overloaded `Is` method on types that's used for shallow and deep comparisons. With #79 raising the need for deep comparisons that allow concrete types to be substituted for DynamicPseudoTypes, it has become obvious we need a clearer interface for this.

I'm proposing:

* `Type.Is(Type)` is a shallow check, that root types are the same type
* `Type.Equals(Type)` is a deep check, that root types and all elements and attributes are the same type
* `Type.ConformsTo(Type)` is a deep check, that root types and all elements and attributes conform (are equivalent, or a concrete type is being substituted for `DynamicPseudoType`)

This would lead to less confusion and fewer gotchas around `Type.Is(Type)`, I believe, and more clarity about what code is actually doing.

bflad:
Closes #80 

Each type now implements three separate and distinct comparison methods:

* `(T).Equal()`: performs deep type equality checks, including attribute/element types and whether attributes are optional or not.
* `(T).Is()`: performs shallow type equality checks, in that the root type is compared, but underlying attribute/element types are not.
* `(T).UsableAs()`: performs type conformance checks. This primarily checks if the target implements `DynamicPsuedoType` in a compatible manner.


paddycarver (approved):
This looks right to me. Really pleased with how this all came out, really nice work.

I was wrong, you _can_ have `null` DynamicPseudoType values. :( We can take care of adding that in another PR, though. I'll open an issue.

PR #99, opened 2021-08-10
tftypes: Allow null DynamicPseudoType in newValue and prevent potential msgpack panic

### LINKED ISSUE #96: tftypes: Disallow msgpack unmarshalling to call NewValue(DynamicPseudoType, nil)

main branch after #94 is reviewed/merged, but should be handled before next release

### Relevant Code

https://github.com/hashicorp/terraform-plugin-go/blob/79438645e8ecca45cf071948cb4b2c647217d60b/tftypes/value_msgpack.go#L321-L326

### Expected Behavior

Code should return error instead of calling `NewValue()` in manner that will panic.

### Actual Behavior

If condition is hit after #94 is merged, code will panic.

### Steps to Reproduce

Needs unit test added.

### References

- #94

### LINKED ISSUE #98: DynamicPseudoType nulls aren't handled correctly

```
post merge of #94
```

### Expected Behavior

DynamicPseudoType values should be able to be null or unknown.

### Actual Behavior

In the wake of #94, DynamicPseudoType values are only able to be unknown.

bflad:
Closes #96
Closes #98 (unreleased bug, no CHANGELOG necessary)


tftypes/primitive.go
```diff
@@ -102,30 +102,11 @@ func (p primitive) supportedGoTypes() []string {
...
-		possibleTypes := []Type{
-			String, Bool, Number,
-			Tuple{}, Object{},
-		}
-		results := []string{}
-		for _, t := range possibleTypes {
-			results = append(results, t.supportedGoTypes()...)
-		}
-		return results
+		return []string{"nil"}
```
paddycarver, line 105: Wouldn't we want `UnknownValue` here, too?

paddycarver (changes requested):

tftypes/primitive.go
bflad: Indeed! Will fix.

```suggestion
		return []string{"nil", "UnknownValue"}
```

paddycarver (approved):

PR #101, opened 2021-08-11
tftypes: Change (AttributePath).WithElementKeyInt() method parameter to int

### LINKED ISSUE #59: QoL:  Make argument to AttributePath.WithElementKeyInt an int (not int64)
### terraform-plugin-go

```
0.2.1
```

### Use cases
When iterating over slices in Go with:
```
for i, v := range mySlice { ... }
```
The type of `i` tends to be `int`. 
Since this seems likely to be the most common source for the value passed into `WithElementKeyInt()` it'd be nice to not have to always cast it into `int64` to make the compiler happy.

Purely from a developer quality of life perspective, I would suggest to make the argument an int.
That is:
```
func (a *AttributePath) WithElementKeyInt(key int) { ... }
```

bflad:
Closes #59


tftypes/attribute_path.go
```diff
@@ -149,7 +149,16 @@ func (a *AttributePath) WithElementKeyString(key string) *AttributePath {
...
+func (a *AttributePath) WithElementKeyInt(key int) *AttributePath {
+	steps := a.Steps()
+	return &AttributePath{
+		steps: append(steps, ElementKeyInt(key)),
+	}
+}
+
+// WithElementKeyInt64 adds an ElementKeyInt step to `a`, using `key` as the
+// element's key. `a` is copied, not modified.
+func (a *AttributePath) WithElementKeyInt64(key int64) *AttributePath {
```
paddycarver, line 161: My controversial hot take is we should leave this out for now, and wait for someone to need it.

paddycarver (changes requested):

tftypes/attribute_path.go
bflad: I'm also completely fine with that. Wasn't sure the boundaries of fully offering "protocol support" (64 bit), but seems pragmatic to drop this here for practical reasons.

paddycarver: I'm generally in the "let's make sure we can support everything the protocol can" camp for this, but I imagine if we have 2 billion elements in a list, something else is going to break first. `cty` _also_ uses an `int`, so it would have problems too. So I'm grudgingly willing to bend to pragmatism on this one. (I also think _Go_ will have some trouble with those values, because I don't know that slices can have more values than `int`s can hold.)

.changelog/101.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:breaking-change
+tftypes: `(AttributePath).WithElementKeyInt()` now has an `int` parameter instead of `int64`.
+```
+
+```release-note:note
```
paddycarver, line 5: Do we need a second note for this, in addition to the breaking change entry?

tftypes/attribute_path.go
bflad: Yeah, the Go spec doesn't support Array or Slice types above `int`: https://golang.org/ref/spec#Array_types

> The length is part of the array's type; it must evaluate to a non-negative constant representable by a value of type int.

It would require a different implementation to always support 64 bit "slices" on non 64 bit systems.

.changelog/101.txt
bflad: Thought it might be nice to explain why the change was made in the changelog, but can remove.

paddycarver (approved):

PR #102, opened 2021-08-16
Update CI for Go 1.17.

paddycarver:


bflad (approved):
Looks good to me 🚀  It may also be worth bumping the `go 1.16` directive in `go.mod` for any tooling based around that value.

PR #105, opened 2021-09-04
The Great Renaming of Map.AttributeType.

paddycarver:
Maps don't have an AttributeType in cty. They have an ElementType. In my
wayward youth of like a year ago, I decided this was A Mistake and in my
great and everlasting wisdom, I changed it to AttributeType. I think my
rationale was that elements are integer-indexed and attributes are
string-indexed.

My hubris has bitten me. This was a bad take. It makes diagnostics weird
(why am I using an ElementKeyString instead of an AttributeName if it's
an attribute?) and deviates with basically no value.

Worse, it's just... wrong. Attributes are structural in nature, like
properties in Go structs. They're defined by the provider developer.
Elements aren't structural, like Go maps. The type has no constraint on
what keys or how many keys can be used. Maps are obviously this second
thing.

This is the great renaming, that will break everything but will finally,
finally correct my blunder.

### DIFF
--- a/.changelog/105.txt
+++ b/.changelog/105.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+The `AttributeType` property of `tftypes.Map` has been renamed to `ElementType`.
+```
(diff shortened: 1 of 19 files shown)


bflad (approved):
Makes sense to me 🚀

PR #109, opened 2021-09-22
Protocol version 6.1

kmoe:
Needs https://github.com/hashicorp/terraform/pull/29572 to be merged so we have the final protobuf file.

Added the new file as `tfplugin6.proto` and ran `./tfprotov6/internal/tfplugin6/generate.sh` to regenerate `.pb.go` bindings.


kmoe:
https://github.com/hashicorp/terraform/pull/29572 is closed - we're taking another approach.

PR #112, opened 2021-09-30
tftypes: Added AttributePath LastStep method

### LINKED ISSUE #111: tftypes: Consider (AttributePath).LastPath() Method

```
v0.4.0
```

### Use cases

When creating `tftypes.Transform()` functions, it can be necessary to know the definition of the last attribute path, to determine the methodology for upserting values into the parent value to replace it.

### Attempted solutions

This code works, although is verbose:

```go
var childValueDiags diag.Diagnostics
childStep := path.Steps()[len(path.Steps())-1] // <-- this here
parentValue, childValueDiags = upsertChildValue(ctx, parentPath, parentValue, childStep, tfVal)
diags.Append(childValueDiags...)
```

### Proposal

Create a method, opposite of [`(AttributePath).WithoutLastStep()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tftypes#AttributePath.WithoutLastStep) that returns the last step, e.g.

```go
func (p *AttributePath) LastStep() *AttributePath
```

### References

- https://github.com/hashicorp/terraform-plugin-framework/blob/2cedf45c2bf94bfb82ff65b1402b3fd75a806aa4/tfsdk/plan.go#L266-L269
- https://github.com/hashicorp/terraform-plugin-framework/blob/2cedf45c2bf94bfb82ff65b1402b3fd75a806aa4/tfsdk/state.go#L275-L278

bflad:
Closes #111

### DIFF
--- a/tftypes/attribute_path_test.go
+++ b/tftypes/attribute_path_test.go
@@ -523,6 +523,62 @@ func TestAttributePathEqual(t *testing.T) {
 	}
 }
 
+func TestAttributePathLastStep(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		path     *AttributePath
+		expected AttributePathStep
+	}{
+		"empty": {
+			path:     NewAttributePath(),
+			expected: nil,
+		},
+		"nil": {
+			path:     nil,
+			expected: nil,
+		},
+		"AttributeName": {
+			path:     NewAttributePath().WithAttributeName("testing"),
+			expected: AttributeName("testing"),
+		},
+		"AttributeName-AttributeName": {
+			path:     NewAttributePath().WithAttributeName("testing").WithAttributeName("testing2"),
+			expected: AttributeName("testing2"),
+		},
+		"AttributeName-AttributeName-AttributeName": {
+			path:     NewAttributePath().WithElementKeyString("testing").WithAttributeName("testing2").WithAttributeName("testing3"),
+			expected: AttributeName("testing3"),
+		},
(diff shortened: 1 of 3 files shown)


tftypes/attribute_path_test.go
```diff
@@ -585,3 +641,38 @@ func TestAttributePathString(t *testing.T) {
 		})
 	}
 }
+
+func attributePathStepComparer(i, j AttributePathStep) bool {
```
paddycarver, line 645: I've been trying to implement an `Equal` method on these that `cmp` will pick up automatically, rather than defining comparers for them. Mostly because the comparers will need to be explicitly included everywhere, and it's hard to keep track of which you need. `Equal` works everywhere without needing to be explicitly included.

paddycarver (changes requested):
A small change to some supporting code, but otherwise seems reasonable to me.

tftypes/attribute_path_test.go
bflad: Good to know that we're generally okay with expanding our compatibility surface area with `Equal()` methods. I think its low risk, but didn't want to presume.

paddycarver: I think in this case it was a no-brainer to me because the interface was intentionally unfulfillable and was less of an interface and more of an enum. So consumers of the package have no real impact as a result of this change.

For other things I give it a bit more consideration, but even then I usually end up siding with including it (see the framework's `attr.Value` and `attr.Type`). No hard and fast rules, unfortunately. :(

Issue #115, opened 2021-11-07
DynamicValue.Unmarshal panics with a null value in DynamicPseudoType from version 0.4.0


iwarapter:
### terraform-plugin-go version

```
github.com/hashicorp/terraform-plugin-go v0.4.0
```

### Relevant provider source code

Previously in version 0.3.1 of plugin-go you could unmarshal a MsgPack for a `DynamicPseudoType` and handle whether a null value was set by the user and return appropriate errors. Now the provider will panic and die.

These two playgrounds show the behaviour for versions 0.3.1 and 0.4.0 with an example MsgPack i took from a unit-test I have.

https://play.golang.org/p/UXllmjPZSHi
Produces:
```
panic: AttributeName("description"): invalid value tftypes.DynamicPseudoType<null> for tftypes.DynamicPseudoType

goroutine 1 [running]:
github.com/hashicorp/terraform-plugin-go/tftypes.NewValue(...)
	/tmp/gopath2210971451/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.4.0/tftypes/value.go:278
github.com/hashicorp/terraform-plugin-go/tftypes.msgpackUnmarshalObject(0xc00010c990, 0xc00010c930, 0xc00011a078)
	/tmp/gopath2210971451/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.4.0/tftypes/value_msgpack.go:331 +0x6d4
github.com/hashicorp/terraform-plugin-go/tftypes.msgpackUnmarshal(0xc0001000c0, {0x54d630, 0xc00010c990}, 0x506c20)
	/tmp/gopath2210971451/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.4.0/tftypes/value_msgpack.go:129 +0x111b
github.com/hashicorp/terraform-plugin-go/tftypes.msgpackUnmarshalDynamic(0xc00010c3c0, 0x54d720)
	/tmp/gopath2210971451/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.4.0/tftypes/value_msgpack.go:357 +0x1d7
github.com/hashicorp/terraform-plugin-go/tftypes.msgpackUnmarshal(0x509800, {0x54d720, 0xc00010c3c0}, 0xd)
	/tmp/gopath2210971451/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.4.0/tftypes/value_msgpack.go:50 +0x11c5
[...]

PR #116, opened 2021-11-11
Accept objects with DynamicPseudoType with null values

kjagiello:
The value of a `DynamicPseudoType` should be allowed to be either unknown or null. The check for the latter was missing and this PR adds it. Let me know if there are any more tests needed.

Solves #115

### DIFF
--- a/.changelog/116.txt
+++ b/.changelog/116.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fixed a panic when using DynamicPseudoType attributes in an object.
+```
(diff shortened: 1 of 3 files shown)


paddycarver:
Hey @kjagiello! Thanks for the PR. :) Any chance you wouldn't mind adding a changelog entry for this? You can do that by including something like

~~~
```release-note:bug
Fixed a panic when using DynamicPseudoType attributes in an object.
```
~~~

Thanks!

paddycarver (approved):

paddycarver:
Thank you!

Issue #107, opened 2021-09-16
Consider Subsystem Logger with Protocol Request and Response Data


bflad:
### terraform-plugin-go version

```
v0.4.0 (not yet released)
```

### Use cases

Provider and Terraform developers may wish to see the request and response data being passed over gRPC during troubleshooting.

### Attempted solutions

Manually replacing the terraform-plugin-go dependency and injecting logging statements.

### Proposal

Create a new logging subsystem specifically for protocol request and response data at `TRACE` log level. This subsystem should default to `OFF` and should be explicitly enabled by the executor, rather infer the log level like other logging subsystems, to prevent leaking sensitive information. The details of this logging subsystem should be documented along side other logging details for this project.

### References

- https://github.com/hashicorp/terraform-plugin-go/pull/93#discussion_r709346132

bflad:
Quick msgpack notes:

```go
b, _ := msgpack.Marshal(/* ... */)
fmt.Printf("Bytes: %v", b) // [142 161 98 207 0 0 0 0 0 0 0 5 161 100 ...

var ret interface{}
decoder := msgpack.NewDecoder(bytes.NewReader(b))
decoder.UseDecodeInterfaceLoose(true)
err := decoder.Decode(&ret)
fmt.Printf("Friendlier: %v", ret) // map[a:5 b:5 c:5 d:true e:foo f:[102 111 111] ...
```

bflad:
Since the data is a little unwieldy, maybe this could dump files separately from the logging. e.g. `{RPC}_(request|response)_{TIMESTAMP}.msgpack` then those could be inspected.

Some additional references for inspection tooling:

- https://github.com/tagomoris/msgpack-inspect
- https://pkg.go.dev/github.com/nokute78/msgpack-microscope
- https://pkg.go.dev/github.com/nokute78/msgpack-microscope/cmd/msgpack2json (and `jq`)
- https://github.com/wader/fq/issues/25

bflad:
To flush out the file-based proposal a little more:

Support a new environment variable, `TF_LOG_SDK_PROTO_DATA_DIR`, which sets a directory to save request and response `DynamicValue` data. When set and a RPC request or response with `DynamicValue` data crosses the server logic, open a file and save the raw data:

- If len(msgpack) > 0, `{TIMESTAMP}_{RPC}_(Request|Response|{MESSAGE})_{FIELD}.msgpack`
[...]

PR #118, opened 2021-12-17
deps: Bump github.com/vmihailenco/msgpack/v4@v4.3.12

bflad:
Reference: https://github.com/vmihailenco/msgpack/compare/v4.0.4...v4.3.12
Reference: https://github.com/hashicorp/terraform/blob/2c8edfb2593a0c3cec044c1504bf1c1edd3f4a94/go.mod#L176

Migrates to Go modules compatible versioning and matches Terraform CLI dependency. ~Was not previously caught by dependabot due to the incompatible versioning.~ dependabot is not enabled. 😦 

Updated via:

```
go get github.com/vmihailenco/msgpack/v4@v4.3.12
# Update import references to include /v4
go mod tidy
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,8 +12,7 @@ require (
 	github.com/hashicorp/terraform-registry-address v0.0.0-20210412075316-9b2996cce896
 	github.com/mitchellh/go-testing-interface v1.14.1
 	github.com/nsf/jsondiff v0.0.0-20200515183724-f29ed568f4ce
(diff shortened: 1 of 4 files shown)


bflad:
@alexsomesan I'm fairly certain this will not resolve your situation, however I'd like to use it as a first step to ensure compatibility and get the latest v4 capabilities of the dependency for #107

paddycarver (approved):

PR #137, opened 2022-01-07
Implement managed debug mode for tf5server and tf6server

### LINKED ISSUE #134: Support/Document Debugger Processes

```
v0.5.0
```

### Use cases

Downstream abstractions, such as terraform-plugin-sdk/v2 and terraform-plugin-framework, make it relatively straightforward to run servers that invert the process lifecycle so they can be instrumented by debugger and testing processes. Providers written directly on terraform-plugin-go or reliant on terraform-plugin-mux essentially must reimplement those projects' logic to do the same. Given that the reattach handling is strongly correlated to low level details about go-plugin and Terraform CLI, it feels like a potential candidate for creating an implementation in this project.

### Proposal

Adding `DebugServe` and/or `Debug` functionality in tf5server/tf6server, similar to downstream implementations. Document in README (until #129 is sorted).

### References

- https://www.terraform.io/plugin/sdkv2/debugging
- https://github.com/hashicorp/terraform-plugin-framework/pull/243
- https://github.com/hashicorp/terraform-plugin-mux/issues/34
- https://github.com/hashicorp/terraform-provider-kubernetes/blob/5808075fda984cdd8e26aacb4cef3490ac542f1e/manifest/provider/plugin.go

bflad:
Closes #134

This will allow providers and downstream SDK implementations to opt into a fully managed debug server, which manages reattach configuration handling and stop signals, rather than requiring duplicate logic that can be unclear and/or verbose to implement.

The `tf5server` and `tf6server` implementations could use some refactoring to reduce _their_ duplicated code (e.g. a shared internal package), however that is left for future handling, especially with other in flight changes at the moment.


bflad:
Verification notes (ensuring providers could be debugged successfully via VS Code):

* In terraform-provider-kubernetes, a muxed provider, the main function debug implementation simplified to only:

```go
if *debugFlag {
	opts = append(opts, tf5server.WithManagedDebug())
}
```

* In terraform-plugin-framework, the new `Debug` flag implementation (https://github.com/hashicorp/terraform-plugin-framework/pull/243) simplifies to the below and bflad/terraform-provider-framework debugging works just the same:

```go
// Serve serves a provider, blocking until the context is canceled.
func Serve(ctx context.Context, providerFunc func() Provider, opts ServeOpts) error {
	var tf6serverOpts []tf6server.ServeOpt

	if opts.Debug {
		tf6serverOpts = append(tf6serverOpts, tf6server.WithManagedDebug())
	}

	return tf6server.Serve(opts.Name, func() tfprotov6.ProviderServer {
		return &server{
			p: providerFunc(),
		}
	}, tf6serverOpts...)
}
```

* In terraform-plugin-sdk, it would require some slight rework before being able to reduce the code, so didn't spend time there.

README.md
```diff
@@ -100,6 +100,25 @@ func TestAccDataSourceFoo(t *testing.T) {
...
+
+```go
+func main() {
+	debugFlag := flag.Bool("debug", false, "Start provider in debug mode.")
+	flag.Parse()
+
+	opts := []tf6server.ServeOpt{}
+
+	if *debugFlag {
+		opts = append(opts, tf5server.WithManagedDebug())
```
ewbankkit, line 115: Should be `tf6server` as the example is for a `tfprotov6` server?

```suggestion
		opts = append(opts, tf6server.WithManagedDebug())
```

bflad: Ugh, yes it should, thank you.

ewbankkit (approved):
LGTM :rocket:.

PR #139, opened 2022-01-14
Increase gRPC maximum message size send and receive limits to 256MB

### LINKED ISSUE #106: Allow configuration of gRPC maximum message sizes

```
v0.3.0
```

### Use cases

The Kubernetes manifest provider has to deal with some potentially huge resource configurations. Particularly, the CRD objects of the  Strimzi project (Apache Kafka on K8s) are as big as 16K (10^3) lines of HCL.

Because of this, it can run into the gRPC message limits quite easily. This is most likely to happen in the `UpgradeResourceState` call first, because the previous state payload is transported in raw JSON for this call which is more verbose than the msgpack payload of other protocol calls.

An instance of this happening would look like this to the user:
```
│ Error: Plugin error
│
│   with kubernetes_manifest.customresourcedefinition_kafkas_kafka_strimzi_io,
│   on crds.tf line 1, in resource "kubernetes_manifest" "customresourcedefinition_kafkas_kafka_strimzi_io":
│    1: resource "kubernetes_manifest" "customresourcedefinition_kafkas_kafka_strimzi_io" {
│
│ The plugin returned an unexpected error from plugin.(*GRPCProvider).UpgradeResourceState: rpc error: code = ResourceExhausted desc = grpc: received message larger than max
│ (4528743 vs. 4194304)

```

### Attempted solutions

I have successfully worked around this issue by maxxing out the message size limits on the gRPC server that terraform-plugin-go instantiates. However, this would require us to consume terraform-plugin-go from a fork and constantly rebase on new releases. This would not be ideal. 

The changes I had to make are available here: https://github.com/alexsomesan/terraform-plugin-go/commit/6f9312ae6c69feb8536cdc7df7057038a3ca32ec

### Proposal

We would prefer for this mechanism be exposed to consumers of terraform-plugin-go so that we can configure practical limits as they are needed.

bflad:
Closes #106

The rationale for choosing this particular value and making it a constant, rather than configurable, is included in the code.

### DIFF
--- a/.changelog/139.txt
+++ b/.changelog/139.txt
@@ -1,0 +1,7 @@
+```release-note:enhancement
+tfprotov5/tf5server: Increased maximum gRPC send and receive message size limit to 256MB
+```
+
+```release-note:enhancement
+tfprotov6/tf6server: Increased maximum gRPC send and receive message size limit to 256MB
+```
--- a/tfprotov5/tf5server/server.go
+++ b/tfprotov5/tf5server/server.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/fromproto"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/toproto"
+	"google.golang.org/grpc"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-plugin"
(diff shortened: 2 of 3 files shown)


alexsomesan (approved):
Thanks for addressing this!

With the introduction of the `kubernetes_manifest` resources, our users have reported running into the previous limit with various configurations, such as the CRDs for Apache Strimzi.

This bump should now enable those use-cases.

Eric-Jckson:
Hey @bflad , 

I have compiled the terraform binary from the main branch and im encountering an error with grpc. Is this error still expected? This is during an update operation.

```
Error: Plugin error
│
│   with module.config-securitypolicy-cep.module.securitypolicy.panos_panorama_security_rule_group.default["Azure"],
│   on modules/terraform-panorama-config-securitypolicy-spw-cep/modules/terraform-panorama-config-securitypolicy-cep/main.tf line 1, in resource "panos_panorama_security_rule_group" "default":
│    1: resource "panos_panorama_security_rule_group" "default" {
│
│ The plugin returned an unexpected error from plugin.(*GRPCProvider).PlanResourceChange: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4460243 vs.
│ 4194304)
```
```
Terraform v1.1.0-dev
on darwin_amd64
+ provider registry.terraform.io/hashicorp/azurerm v2.93.0
+ provider registry.terraform.io/hashicorp/null v3.1.0
+ provider registry.terraform.io/paloaltonetworks/panos v1.9.0
```

PR #140, opened 2022-01-14
Regenerate Protocol Buffers Go code to remove dependency on deprecated Go module

bflad:
Reference: https://github.com/golang/protobuf/pull/1306

This does not appear to be generating any meaningful differences.

Updated via:

```shell
pushd tfprotov5/internal/tfplugin5
./generate.sh
popd
pushd tfprotov6/internal/tfplugin6
./generate.sh
popd
go mod tidy
```


alexsomesan (approved):
Looks harmless to me :)

PR #150, opened 2022-02-02
Propagate provider name into GRPCProviderPlugin

alexsomesan:
This resolves the bellow error message printed on provider startup:
```
2022/02/02 01:07:48 [ERROR] Error parsing provider name "": Invalid provider source string: The "source" attribute must be in the format "[hostname/][namespace/]name"

```

### DIFF
--- a/tfprotov5/tf5server/server.go
+++ b/tfprotov5/tf5server/server.go
@@ -258,6 +258,7 @@ func Serve(name string, serverFactory func() tfprotov5.ProviderServer, opts ...S
 		},
 		Plugins: plugin.PluginSet{
 			"provider": &GRPCProviderPlugin{
+				Name:         name,
 				GRPCProvider: serverFactory,
 			},
 		},


bflad (approved):
This is a great catch, @alexsomesan! Thank you for fixing this. It looks like we don't have this issue in `tf6server`, so 🎉  there.

Could you also include a [changelog entry](https://github.com/hashicorp/terraform-plugin-go/blob/main/.github/CONTRIBUTING.md#changelog)? Something like the below in `.changelog/150.txt` would be great.

``````markdown
```release-note:bug
tfprotov5/tf5server: Prevent empty provider address error logs on startup
```
``````

Thanks again!

bflad:
I discovered through testing https://github.com/hashicorp/terraform-plugin-sdk/pull/857, we also need to pass `Opts: opts` through there as well, otherwise acceptance testing logging does not constrain the protocol logging by default. I'm going to pull this in (after pushing the changelog entry to your branch) and separately work on that issue, hopefully with some unit testing.

alexsomesan:
Thanks for doing my work for me and adding the change-log entry. Got sidetracked and completely forgot about it.

PR #159, opened 2022-02-24
Fixing attribute path step bug

preved911:
Hello. I'm using [hashicorp/terraform-provider-kubernetes](https://github.com/hashicorp/terraform-provider-kubernetes) provider with `wait_for` field customization. In some cases it will not work correctly. For example, I'm trying to install manifest for [zalando/postgres-operator](https://github.com/zalando/postgres-operator) and specify `wait_for` like this:
```
  wait_for = {
    fields = {
      "status.PostgresClusterStatus" = "Running"
    }
  }
```
In my case, `apply` will not finish even after field `.status.PostgresClusterStatus` is become in value `Running`.
As I understand correct, this patch should fix such behavior.

### DIFF
--- a/tftypes/value.go
+++ b/tftypes/value.go
@@ -133,7 +133,7 @@ func (val Value) ApplyTerraform5AttributePathStep(step AttributePathStep) (inter
 	}
 	switch s := step.(type) {
 	case AttributeName:
-		if !val.Type().Is(Object{}) {
+		if !val.Type().Is(Object{}) && !val.Type().Is(Map{}) {
 			return nil, ErrInvalidStep
 		}
 		o := map[string]Value{}


bflad (changes requested):
Hi @preved911 👋  Thank you for submitting this and sorry you ran into trouble here.

Is there an associated GitHub issue for reproducing the problem so we can verify this change? We would also want to confirm the fix via a unit test in this repository.

My understanding is that value traversal of a map should occur via the `ElementKeyString` step while `AttributeName` is reserved for an object. If that understanding is incorrect, it would be good to see the related provider code as it may be trying to do value traversal in an unexpected manner than we have seen before.

Thanks.

alexsomesan:
@preved911 have you already reported this issue in [hashicorp/terraform-provider-kubernetes](https://github.com/hashicorp/terraform-provider-kubernetes) repository?

preved911:
> @preved911 have you already reported this issue in [hashicorp/terraform-provider-kubernetes](https://github.com/hashicorp/terraform-provider-kubernetes) repository?

No, I didn't do it yet. It's also need?

> If that understanding is incorrect, it would be good to see the related provider code as it may be trying to do value traversal in an unexpected manner than we have seen before.

The problem, which I got, is related with [waiter](https://github.com/hashicorp/terraform-provider-kubernetes/blob/1340e20afebd531828df2e46836afda9bd5f922a/manifest/provider/waiter.go#L112:L190) logic. It can be reproduced with any crd-resource, where field contain [additionalProperties](https://github.com/zalando/postgres-operator/blob/d032e4783eb84fcc9c3056d374f4b6340f18141b/charts/postgres-operator/crds/postgresqls.yaml#L643:L644) spec.

bflad:
Hi again @preved911 👋  My understanding of Kubernetes Custom Resource Definitions is limited, so I may be misunderstanding things, but it appears they follow OpenAPIv3 specifications where an object type may define either `properties` or [`additionalProperties`]().

In this case, the schema definition is:

```go
          status:
            type: object
            additionalProperties:
              type: string
```

In terms of how these object types should be translated into the `tftypes` types, it appears that `properties` should be `Object` (with `AttributeName` steps) since the underlying attributes are well-defined and that `additionalProperties` should be `Map` (with `ElementKeyString` steps) since the key names are unknown. The provider logic may need to be adjusted to step into objects with `additionalProperties` via `ElementKeyString` instead of `AttributeName`.

This distinction between stepping via `AttributeName` versus `ElementKeyString` may seem inconsequential, however it is a design decision to ensure that the data types of values are aligned during access.

@alexsomesan are you able to take a look into the provider code further to see why this might be occurring? Please reach out if you want to meet up to discuss it further.

preved911:
@bflad, thanks for explanation. So, this PR should be closed?

bflad:
We can close it for now because the current implementation matches the design intention. If we find that the design should be modified, we can always revisit this.

I'd suggest following up by an issue in [terraform-provider-kubernetes](https://github.com/hashicorp/terraform-provider-kubernetes/issues) for further tracking. Thanks, @preved911!

alexsomesan:
@bflad I'll have a look at this on our side. 

I'm fairly certain this is a side-effect of some oversight on our side when we upgraded to plugin-go v0.5.0 with the updated AttributePath indexing elements.

preved911:
UPD: It's my bug. I misunderstanded documentation. Provider work well in specified case.

PR #173, opened 2022-03-22
Add CI Workflow for Protocol Buffers

### LINKED ISSUE #172: Add CI Workflow for Protobuf Compilation

```
v0.8.0
```

### Use cases

The Protocol Buffers compilation via the `protoc` tool is currently handled outside `go generate` so all dependencies of that execution can be handled via Go modules and `go get`. There can be changes to this ecosystem we would like to follow along for updates, so we should raise our confidence in these changes by introducing additional CI processes for this area.

### Attempted solutions

Locally, switch to dependency upgrade branch, install `protoc` (if not done so already), run `protoc`, and check Git for differences.

### Proposal

Introduce a new GitHub Actions based workflow specifically for protobuf changes, e.g. `.github/workflows/ci-protobuf.yml`. That:

* Only has read-only permissions to the repository contents
* Only runs on the `pull_request` event of certain paths such as:
    * `.github/workflows/ci-protobuf.yml`
    * `go.mod`
    * `go.sum`
    * `tfprotov5/internal/tfplugin5/*`
    * `tfprotov6/internal/tfplugin6/*`
* Checks out code
* Installs Go (e.g. actions/setup-go)
* Installs `protoc` (e.g. arduino/setup-protoc)
* Runs `protoc` (refer also to generate.sh)
* Checks Git for differences and fails if any are detected

### References

- https://github.com/hashicorp/terraform-plugin-go/pull/171

bflad:
Closes #172


tools/tools.go
```diff
@@ -0,0 +1,10 @@
+//go:build tools
+// +build tools
```
detro, line 2: Isn't `+build` deprecated in favour of `go:build` now?

.github/workflows/ci-protobuf.yml
```diff
@@ -0,0 +1,47 @@
...
+    paths:
+      - .github/workflows/ci-protobuf.yml
+      - .go-version
+      - go.mod
+      - go.sum
+      - 'tfprotov5/internal/tfplugin5/*'
+      - 'tfprotov6/internal/tfplugin6/*'
+
+permissions:
+  contents: read
```
detro, line 15: And this doesn't apply to the filesystem where the generation below will be executed, correct?

I looked on [the doc](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) but it wasn't specified, so I assume the filesystem is always accessible for write, but then any attempt to persist the code would fail.

Also, setting only `contents` leaves all other permissions on the default level, right? [I couldn't determine this either](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

.github/workflows/ci-protobuf.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  protoc:
```
detro, line 18: So, this is all good, but I think we should add a couple of things so that the knowledge isn't so "buried".

I was looking for a `Makefile` in the project to help doing something like `make generate` (to take care of the code generation). And there is no explanation about this bit of the development workflow in the `README`.

So I dug up the `generate.sh` and found:

```
# We do not run protoc under go:generate because we want to ensure that all
# dependencies of go:generate are "go get"-able for general dev environment
# usability. To compile all protobuf files in this repository, run
# "make protobuf" at the top-level.
```

So, I suspect this is a rethinking of the original approach (that I frankly don't fully understand because it wasn't explained in the `generate.sh` either).

Also, that bit above mentions a `Makefile` that doesn't seem to exist anymore.

I don't come from an "`Makefile`-heavy" background, but I think it would be preferable if all "developer's default actions" (not sure how to best describe it) were ready made in there. Compared to the README, that is textual description, the `Makefile` helps on a more "don't tell me, show me" level.

What do you think?

detro: As a matter of fact, I'd argue that the `generate.sh` scripts should just be replaced with `Makefile` commands, given that all the script does is:

- figure out its own directory
- call `protoc`

.github/workflows/ci-protobuf.yml
bflad: `contents` refers to the REST API endpoints the repository. It's not intuitive, and I'm not sure where that's documented, but its why permissions such as `issues: write` is necessary for modifying milestones, since those APIs are related to issues subsystem.

Assigning one of the permissions removes other permissions per [this documentation](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs):

> If you specify the access for any of these scopes, all of those that are not specified are set to `none`.

Makefile
```diff
@@ -0,0 +1,29 @@
+default: test
+
+lint:
+	golangci-lint run ./...
+
+# Protocol Buffers compilation is done outside of 'go generate' handling since
```
detro, line 6: 👩‍🍳 😘

detro (approved):
This is excellent.

PR #190, opened 2022-05-11
Invalid plan arising from integer rounding

### LINKED ISSUE #189: Invalid plan arising from integer rounding
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.7.0
```

```go
"min": {
	Type:          types.Int64Type,
},
"max": {
	Type:          types.Int64Type,
},
```

### Terraform Configuration Files

```hcl
resource "random_integer" "integer_1" {
   max  = 7227701560655103598
   min  = 7227701560655103597
   seed = 12345
}
```

### Debug Output
https://gist.github.com/bendbennett/6d9066c120883e17dcda25f0e98aec59

### Expected Behavior
There should be no integer rounding.

### Actual Behavior
Rounding is observed which causes an invalid plan.
```
Provider "registry.terraform.io/hashicorp/random" planned an invalid value
        for random_integer.integer_1.min: planned value
        cty.NumberIntVal(7.227701560655104e+18) does not match config value
        cty.NumberIntVal(7.227701560655103597e+18).
```

### Steps to Reproduce
Use a 64 bit integer within a resource that uses Framework v0.7.0

bendbennett:
#### Issue 1
A lossy conversion between int64 and uint64 can arise because of the use of `float64()`, which uses a float64 with a precision of 53.

This issue arises in the following locations:
- [value_msgpack.go#L77](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack.go#L77)
- [value_msgpack.go#L85](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack.go#L85)
- [value_msgpack.go#L91](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack.go#L91)

Taking the form:
```go
return NewValue(Number, big.NewFloat(float64(rv))), nil
```

#### Issue 2
Another potential lossy operation can arise from the usage of `big.Float.Set`.

This issue arises in the following location:
-   [value.go#L443](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value.go#L443)

This is a consequence of the following:
```go
v, ok := val.value.(*big.Float)
if !ok {
	return fmt.Errorf("can't unmarshal %s into %T, expected *big.Float", val.Type(), dst)
}
target.Set(v)
```

Because `v` uses the default precision of 53 of `big.Float`

An additional consequence of using `target.Set(v)` is that the tests within [value_msgpack_test.go](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack_test.go) would not detect this form of rounding.

### Fixes
- `new(big.Float).SetInt64()` or `new(big.Float).SetUint64()` is used as appropriate in [value_msgpack.go](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack.go).
- `big.Float.Copy` instead of `big.Float.Set` is used within [value.go](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value.go).
- Tests within [value_msgpack_test.go](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value_msgpack_test.go) have been updated to check for rounding.

### Related Issues
- Closes #189

### DIFF
--- a/tftypes/value.go
+++ b/tftypes/value.go
@@ -440,7 +440,7 @@ func (val Value) As(dst interface{}) error {
 		if !ok {
 			return fmt.Errorf("can't unmarshal %s into %T, expected *big.Float", val.Type(), dst)
 		}
-		target.Set(v)
+		target.Copy(v)
 		return nil
 	case **big.Float:
 		if val.IsNull() {
--- a/tftypes/value_msgpack.go
+++ b/tftypes/value_msgpack.go
@@ -74,7 +74,7 @@ func msgpackUnmarshal(dec *msgpack.Decoder, typ Type, path *AttributePath) (Valu
 			if err != nil {
 				return Value{}, path.NewErrorf("couldn't decode number as int64: %w", err)
 			}
-			return NewValue(Number, big.NewFloat(float64(rv))), nil
+			return NewValue(Number, new(big.Float).SetInt64(rv)), nil
 		}
 		switch peek {
 		case msgpackCodes.Int8, msgpackCodes.Int16, msgpackCodes.Int32, msgpackCodes.Int64:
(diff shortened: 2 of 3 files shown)


detro (approved):
👨‍🍳 😘

bflad (approved):
Looks good to me, just needs a CHANGELOG entry added, e.g. in `.changelog/190.txt`

``````
```release-note:bug
tftypes: Prevented loss of number precision with integers between 54 and 64 bits
```
``````

PR #200, opened 2022-06-13
Switch to the new minimal terraform-registry-address API

radeksimko:
Depends on https://github.com/hashicorp/terraform-registry-address/pull/12

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -44,6 +44,7 @@ jobs:
       - run: go mod edit -replace github.com/hashicorp/terraform-plugin-go=../
       - run: go mod tidy
       - run: go test -v ./internal/protocolprovider
+      - run: go test -v ./internal/framework5provider
       - run: go test -v ./internal/sdkv2provider
         env:
           TF_ACC: '1'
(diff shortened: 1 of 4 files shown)


radeksimko:
Not sure what the CI failures are about - they look unrelated to this PR - let me know if they are, I can take a deeper look.

detro (commented):
LGTM

detro:
Mmm, I think the paths used when running `terraform-provider-corner` tests need to be updated. Let me see if I can find an example PR.

detro:
Here are the necessary changes: https://github.com/hashicorp/terraform-plugin-framework/pull/385/files

radeksimko:
@detro I addressed the failures by adjusting the paths in the workflow file which seemed sufficient to get it all green.

PTAL

bflad (approved):
Looks good to me 🚀

bflad:
@radeksimko are you going to cut a tagged release first or would you like these merged first? Thanks!

radeksimko:
@bflad I'm planning to merge all the PRs first and possibly still try to see if TF Registry itself could also consume it - which may still affect the API - but I might end up leaving that one for v2 if it gets too hairy.

Issue #212, opened 2022-07-26
Unmarshal should ignore elements in JSON that do not have a corresponding attribute


bendbennett:
### terraform-plugin-go version

```
0.12
```

### Relevant provider source code

```go
func TestValueFromJSON(t *testing.T) {
	t.Parallel()
	type testCase struct {
		value Value
		typ   Type
		json  string
	}
	tests := map[string]testCase{
		"object-with-missing-attribute": {
			value: NewValue(Object{
				AttributeTypes: map[string]Type{
					"bool":   Bool,
					"number": Number,
				},
			}, map[string]Value{
				"bool":   NewValue(Bool, true),
				"number": NewValue(Number, big.NewFloat(0)),
			}),
			typ: Object{
				AttributeTypes: map[string]Type{
					"bool":   Bool,
					"number": Number,
				},
			},
			json: `{"bool":true,"number":0,"unknown":"whatever"}`,
		},
	}
	for name, test := range tests {
		name, test := name, test
		t.Run(name, func(t *testing.T) {
			t.Parallel()
			val, err := ValueFromJSON([]byte(test.json), test.typ)
			if err != nil {
				t.Fatalf("unexpected error unmarshaling: %s", err)
			}
			if diff := cmp.Diff(test.value, val); diff != "" {
				t.Errorf("Unexpected results (-wanted +got): %s", diff)
			}
		})
	}
}
```

### Expected Behavior
The test should pass.

### Actual Behavior
The test fails with the following error:
```
unexpected error unmarshaling: ElementKeyValue(tftypes.String<unknown>): unsupported attribute "unknown"
```

### Proposals
- Modify the behaviour of `Unmarshal` so that elements that do not have a corresponding attribute can optionally be silently ignored.  

### References
This behaviour was discovered during the investigation of [Unable to Read Previously Saved State for UpgradeResourceState TF 0.12](https://github.com/hashicorp/terraform-plugin-framework/issues/415).

PR #213, opened 2022-07-27
Allow ignoring undefined attributes during unmarshalling

bendbennett:
Closes: #212 

An error is generated when unmarshalling JSON that contains fields that do not have a corresponding attribute.

This PR adds an additional function, `UnmarshalWithOpts` that allows supplying of options that modify unmarshalling behaviour. Specifically, ignoring of undefined attributes is configurable.

### DIFF
--- a/.changelog/213.txt
+++ b/.changelog/213.txt
@@ -1,7 +1,7 @@
 ```release-note:enhancement
-tfprotov5/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+tfprotov5: Added `RawState` type `UnmarshalWithOpts` method to facilitate configurable behaviour during unmarshalling
 ```
 
 ```release-note:enhancement
-tfprotov6/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+tfprotov6: Added `RawState` type `UnmarshalWithOpts` method to facilitate configurable behaviour during unmarshalling
 ```
--- a/.changelog/213.txt
+++ b/.changelog/213.txt
@@ -1,0 +1,7 @@
+```release-note:enhancement
+tfprotov5/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+```
+
+```release-note:enhancement
+tfprotov6/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+```
--- a/tfprotov5/state.go
+++ b/tfprotov5/state.go
@@ -82,7 +82,7 @@ func (s RawState) Unmarshal(typ tftypes.Type) (tftypes.Value, error) {
 // options that can be used to modify the behaviour when unmarshalling JSON or Flatmap.
 func (s RawState) UnmarshalWithOpts(typ tftypes.Type, opts tftypes.UnmarshalOpts) (tftypes.Value, error) {
 	if s.JSON != nil {
-		return tftypes.ValueFromJSONWithOpts(s.JSON, typ, opts.JsonOpts) //nolint:staticcheck
+		return tftypes.ValueFromJSONWithOpts(s.JSON, typ, opts.JSONOpts) //nolint:staticcheck
 	}
 	if s.Flatmap != nil {
 		return tftypes.Value{}, fmt.Errorf("flatmap states cannot be unmarshaled, only states written by Terraform 0.12 and higher can be unmarshaled")
--- a/tfprotov5/state.go
+++ b/tfprotov5/state.go
@@ -77,3 +77,15 @@ func (s RawState) Unmarshal(typ tftypes.Type) (tftypes.Value, error) {
 	}
 	return tftypes.Value{}, ErrUnknownRawStateType
 }
+
+// UnmarshalWithOpts is identical to Unmarshal but also accepts a tftypes.UnmarshalOpts which contains
+// options that can be used to modify the behaviour when unmarshalling JSON or Flatmap.
+func (s RawState) UnmarshalWithOpts(typ tftypes.Type, opts tftypes.UnmarshalOpts) (tftypes.Value, error) {
+	if s.JSON != nil {
+		return tftypes.ValueFromJSONWithOpts(s.JSON, typ, opts.JsonOpts) //nolint:staticcheck
+	}
+	if s.Flatmap != nil {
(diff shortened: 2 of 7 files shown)


tftypes/value_json.go
```diff
@@ -16,7 +16,28 @@ import (
...
 // not covered under the API compatibility guarantees. Don't use this.
 func ValueFromJSON(data []byte, typ Type) (Value, error) {
-	return jsonUnmarshal(data, typ, NewAttributePath())
+	return jsonUnmarshal(data, typ, NewAttributePath(), JSONOpts{})
+}
+
+// UnmarshalOpts contains options that can be used to modify the behaviour when
+// unmarshalling. Currently, this only contains a struct for opts for JSON but
+// could have a field for Flatmap in the future.
+type UnmarshalOpts struct {
```
bflad, line 25: 😭 Worst nit ever... could we move this to both the tfprotov5 and tfprotov6 packages so its directly tied with the `Unmarshal` function its providing options for? I'm _very_ doubtful we would have differing implementations between the two protocol versions, but I'd hate to be wrong and its hopefully not an expensive choice to keep them in sync as necessary.

tftypes/value_json.go
```diff
@@ -16,7 +16,28 @@ import (
...
+	JSONOpts JSONOpts
+}
+
+// JSONOpts contains options that can be used to modify the behaviour when
+// unmarshalling JSON.
+type JSONOpts struct {
```
bflad, line 31: Nit: Can we name this `ValueFromJSONOpts` to match the `ValueFromJSON` function it is associated with? I'm sorry this is pedantic but there's other places that deal with JSON and it could be confusing if we ever needed another JSON options struct like this.

tftypes/value_json.go
```diff
@@ -422,7 +442,11 @@ func jsonUnmarshalTuple(buf []byte, elementTypes []Type, p *AttributePath) (Valu
 	}, vals), nil
 }
 
-func jsonUnmarshalObject(buf []byte, attrTypes map[string]Type, p *AttributePath) (Value, error) {
+// jsonUnmarshalObject attempts to decode JSON object structure to tftypes.Value object.
+// opts contains fields that can be used to modify the behaviour of JSON unmarshalling.
+// ignoreUndefinedAttributes is used to ignore any attributes which appear in the JSON but do not have a corresponding
+// entry in the schema. For example, raw state where an attribute has been removed from the schema.
```
bflad, line 448: Nit: Can you move this up to the Go documentation for the options field? Thanks!

tftypes/value_json_test.go
```diff
@@ -380,3 +380,47 @@ func TestValueFromJSON(t *testing.T) {
...
+
+func TestValueFromJSONWithOpts(t *testing.T) {
+	t.Parallel()
+	type testCase struct {
+		value Value
+		typ   Type
+		json  string
+	}
+	tests := map[string]testCase{
+		"object-with-missing-attribute": {
```
bflad, line 392: Nit: Can you include a "happy" path case as well, where it doesn't ignore any attributes?

tfprotov5/state.go
```diff
@@ -77,3 +77,15 @@ func (s RawState) Unmarshal(typ tftypes.Type) (tftypes.Value, error) {
 	}
 	return tftypes.Value{}, ErrUnknownRawStateType
 }
+
+// UnmarshalWithOpts is identical to Unmarshal but also accepts a tftypes.UnmarshalOpts which contains
+// options that can be used to modify the behaviour when unmarshalling JSON or Flatmap.
+func (s RawState) UnmarshalWithOpts(typ tftypes.Type, opts tftypes.UnmarshalOpts) (tftypes.Value, error) {
```
bflad, line 83: Could you add brief unit tests for these new `UnmarshalWithOpts` functions (similar to the `ValueFromJSONWithOpts` ones)?

We'll also need to ensure all these new pieces are captured in a CHANGELOG entry file. 👍 

Thanks!

bflad (commented):
Overall this looks really good -- just a few minor things to tidy up and I think we'll be good to go!

tftypes/value_json.go
bendbennett: Makes sense. Have moved.

tftypes/value_json.go
bendbennett: Consistency is good :)
Have renamed.

tftypes/value_json.go
bendbennett: Moved.

tftypes/value_json_test.go
bendbennett: Included "happy" path.

tfprotov5/state.go
bendbennett: Added unit tests for `UnmarshalWithOpts`.

.changelog/213.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:enhancement
+tfprotov5/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+```
+
+```release-note:enhancement
+tfprotov6/state: Added `UnmarshalWithOpts` func to facilitate configurable behaviour during unmarshalling
+```
```
bflad, line 7:
``````suggestion
```release-note:enhancement
tfprotov5: Added `RawState` type `UnmarshalWithOpts` method to facilitate configurable behaviour during unmarshalling
```

```release-note:enhancement
tfprotov6: Added `RawState` type `UnmarshalWithOpts` method to facilitate configurable behaviour during unmarshalling
```
``````

File: tftypes/value_json.go [hunk 7]
```diff
@@ -16,7 +16,25 @@ import (
...
+}
+
+// ValueFromJSONOpts contains options that can be used to modify the behaviour when
+// unmarshalling JSON.
+//
+// IgnoreUndefinedAttributes is used to ignore any attributes which appear in the
+// JSON but do not have a corresponding entry in the schema. For example, raw state
+// where an attribute has been removed from the schema.
+type ValueFromJSONOpts struct {
+	IgnoreUndefinedAttributes bool
```
bflad, line 29: Nit: If you move the field documentation above the attribute, Go docs will show that text when linked directly to the field 😄 

```suggestion
type ValueFromJSONOpts struct {
	// IgnoreUndefinedAttributes is used to ignore any attributes which appear in the
	// JSON but do not have a corresponding entry in the schema. For example, raw state
	// where an attribute has been removed from the schema.
	IgnoreUndefinedAttributes bool
```

bflad (approved):
Looks great! Just two little bitty things to consider 🚀

PR #214, opened 2022-07-27
tftypes: Clarified `ValueFromJSON` error messaging with object attribute key issues

### LINKED ISSUE #211: Unexpected ElementKeyValue Error Output from (tfprotov#.RawState).Unmarshal with Undefined Attributes

```
v0.12.0
```

```go
_, err := req.RawState.Unmarshal(ctx)

if err != nil {
  tfsdklog.Error(ctx, "Oh no!", map[string]interface{}{"error": err})
}
```

### Expected Behavior

The error messaging includes `AttributeName`, to indicate a step into an object type, or does not include incorrect path information.

### Actual Behavior

The error messaging currently includes `KeyElementValue`, which is supposed to indicate a step into a set type.

For example, a `RawState` that includes an undefined root-level attribute, `example`, will return an error message such as:

```text
ElementKeyValue(tftypes.String<unknown>): unsupported attribute "example"
```

This is especially confusing when the undefined attribute is not a string.

The object decoding code initially tries to read the attribute name string key, by setting the "current" attribute path like this to indicate its unknown, but it feels more confusing to say the wrong thing here, rather than just omitting that non-existent and incorrect attribute path information or fixing it to indicate a correct "current" step.

### Proposals

We could consider between either:

- Wholly dropping the "current" step `ElementKeyValue(tftypes.String<unknown>)` path information (which could have other path in front of it if nested)
- Dropping the "current" step `ElementKeyValue(tftypes.String<unknown>)` path information, but additionally handling the special case of it being a root-level attribute that is missing by checking the path length for 0 and returning a synthetic `Root: unsupported attribute %q` error in that case instead.
- Setting the "current" step to `AttributeName(key)` so the error output will include that instead

### References

https://github.com/hashicorp/terraform-plugin-go/blob/4a24b5d5377f223ca890cdada0b6fdbab7410d45/tftypes/value_json.go#L438-L451

bflad:
Closes #211


bendbennett (approved):
LGTM

PR #221, opened 2022-08-12
tfprotov5/tf5server+tfprotov6/tf6server: Include private state data in protocol data output

### LINKED ISSUE #220: Include Resource Private State In Protocol Data Output

```
v0.13.0
```

### Use cases

Using the `TF_LOG_SDK_PROTO_DATA_DIR` environment variable, it is possible to output the data being sent between Terraform and providers. This data does not currently include the resource private state, which is managed solely by providers, but must be passed across the protocol to be preserved. It can be helpful to see this data when troubleshooting private state functionality.

### Attempted solutions

Viewing private state information via the state file from manual Terraform executions, outside the provider acceptance testing framework.

### Proposal

For protocol RPCs that have private state fields, include them in `TF_LOG_SDK_PROTO_DATA_DIR` output:

* If empty, writing a `..._{FIELD}.empty` file.
* If non-empty, writing the `[]byte` directly to a `..._{FIELD}` file.

### References

- https://github.com/hashicorp/terraform-plugin-framework/pull/452

bflad:
Closes #220

This data is "private" in the sense that it is provider-owned, rather than something managed by Terraform.


bendbennett (approved):
Nice. LGTM

PR #231, opened 2022-10-13
tfprotov5+tfprotov6: Synchronize Protocol Buffers comments for ReadResource and UpgradeResourceState

bflad:
Reference: https://github.com/hashicorp/terraform/pull/31998


bendbennett (approved):
LGTM

PR #258, opened 2023-02-03
Add Changie changelog automation

SBGoods:


.github/workflows/release.yml
```diff
@@ -1,9 +1,16 @@
...
+  workflow_dispatch:
+    inputs:
+      versionNumber:
+        description: 'Release version number'
+        type: string
+        required: true
+
+env:
+  CI_COMMIT_AUTHOR: hc-github-team-tf-provider-devex
+  CI_COMMIT_EMAIL: github-team-tf-provider-devex@hashicorp.com
```
bflad, line 13: We should consider moving these to [configuration variables](https://github.blog/changelog/2023-01-10-github-actions-support-for-configuration-variables-in-workflows/) in the future.

.github/workflows/release.yml
```diff
@@ -1,9 +1,16 @@
 name: release
 
 on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+*'
+  workflow_dispatch:
+    inputs:
+      versionNumber:
+        description: 'Release version number'
```
bflad, line 7: Should this let folks know that it should include the `v` prefix?

```suggestion
        description: 'Release version number (v#.#.#)'
```

bflad (approved):

.github/workflows/release.yml
SBGoods: Created an [issue](https://github.com/hashicorp/terraform-providers-devex-internal/issues/129) to track this.

PR #266, opened 2023-02-28
all: Bump minimum Go version to 1.19

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-go/issues/265

Bumped via:

```shell
go mod edit -go=1.19
go mod tidy
go fix ./...
```


bflad:
I was able to sort out my local generation issue. I had run `brew install protoc-gen-go` back on Go 1.18 and since brew builds from source, so it was using that older Go version comment formatting. `brew reinstall protoc-gen-go` did the trick.

This is ready for review and will handle #268 afterwards (Homebrew already has the new version).

bendbennett (approved):
LGTM 👍

PR #280, opened 2023-03-22
deps: Upgrade to github.com/vmihailenco/msgpack/v5

bflad:
Reference: https://github.com/vmihailenco/msgpack/compare/v4.3.12...v5.3.5

The v4 branch is no longer maintained, the v5 branch is stable, and v4 declares `google.golang.org/appengine` amongst some others as dependencies, which are not used by our code and cannot be pruned from this Go module's dependency graph.


austinvalle (approved):
🤷🏻  -> 👍🏻

PR #292, opened 2023-06-05
[COMPLIANCE] Add copyright and license headers

austinvalle:
Closes #259 

This PR introduces:
- Add new `tools` module, add to dependabot, and move GRPC program to `tools` module dependencies
- `make generate` command that will auto-generate all copyright headers for this repo (hooks into `go generate`)
- `make tools` command to simplify the installation process of the Go specific tooling for `make protoc` target
    - Updated GH workflow to use this `make tools` command
- GH workflow that will error if copyright headers aren't present or if license is malformed
- Addition to the `CONTRIBUTING.md` on how to run license autogenerating commands


.copywrite.hcl
```diff
@@ -0,0 +1,31 @@
...
+    # golangci-lint tooling configuration
+    ".golangci.yml",
+
+    # GoReleaser tooling configuration
+    ".goreleaser.yml",
+
+    # Release Engineering tooling configuration
+    ".release/*.hcl",
+
+    # Proto and protoc generated files
```
austinvalle, line 27: Along with the auto-generated files, I also ignored the `.proto` file. I'm assuming that is okay? I believe comments would persist to the auto-generated file as well if we wanted to just ignore the auto-generated files only, i.e. `*.pb.go`

Makefile
```diff
@@ -3,6 +3,10 @@ default: test
 lint:
 	golangci-lint run ./...
 
+tools:
+	go install google.golang.org/protobuf/cmd/protoc-gen-go
```
bflad, line 7: It's a bummer this cannot also be moved (e.g. imports are baked into the `.pb.go` generated files) 😄

tools/tools.go
```diff
@@ -1,10 +1,17 @@
-//go:build tools
-// +build tools
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+//go:build generate
```
bflad, line 4: ❤️

.copywrite.hcl
```diff
@@ -0,0 +1,31 @@
...
+
+    # GoReleaser tooling configuration
+    ".goreleaser.yml",
+
+    # Release Engineering tooling configuration
+    ".release/*.hcl",
+
+    # Protoc generated files
+    "tfprotov5/internal/tfplugin5/*.pb.go",
+    "tfprotov6/internal/tfplugin6/*.pb.go",
```
bflad, line 29: Hopefully `copywrite` should be happier now with the headers being included after auto-generation, but if not, let's followup after this PR. 👍 

```suggestion
```

bflad (approved):
Looks good to me 🚀

bflad (approved):
Looks good to me 🚀

.copywrite.hcl
austinvalle: Removed ⚡

Makefile
austinvalle: Agreed 😞

PR #295, opened 2023-06-08
ci: Update arduino/setup-protoc to v2.0.0

bflad:
Pinning to latest workflow major version, which supports recent `protoc` version updates after upstream release changes and runs on NodeJS 16 which removes that deprecation warning.


austinvalle (approved):
awesome! 🚀

PR #301, opened 2023-06-16
Fix StopProvider RPC function call in v6

austinvalle:
Ref: https://github.com/hashicorp/terraform-plugin-framework/issues/781


bflad (approved):
Looks good to me 🚀

PR #306, opened 2023-06-27
tfprotov5+tfprotov6: Add DynamicValue type IsNull method

### LINKED ISSUE #305: tfprotov5+tfprotov6: Add DynamicValue Null Checking Without Type Information

```
v0.16.0
```

### Use cases

terraform-plugin-mux, `tfprotov5.ProviderServer`, and `tfprotov6.ProviderServer` implementations generally need to check whether a `DynamicValue` is null. For example, this type of check can be used in `PlanResourceChange` RPCs to verify whether the intended resource operation is creation or deletion. Currently, `DynamicValue` only supports an `Unmarshal()` method, which must be given the expected type information (e.g. from a schema), which may not always be available immediately in implementations.

### Attempted solutions

Caching schema information, calling `(DynamicValue).Unmarshal()`, then checking `(tftypes.Value).IsNull()`.

### Proposal

Checking whether a `DynamicValue` is null can be performed by peeking into the underlying JSON and MessagePack data without fully decoding it. JSON data should have a `null` value (`nil` `json.Token`) and MessagePack data should have a `msgpcode.Nil` code.

In both `tfprotov5` and `tfprotov6` packages, add a new `IsNull()` method to the `DynamicValue` type.

### References

- https://github.com/hashicorp/terraform-provider-aws/issues/31722
- https://github.com/hashicorp/terraform-plugin-mux/issues/168

bflad:
Closes #305

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20230627-125806.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20230627-125806.yaml
@@ -1,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'tfprotov5: Added `DynamicValue` type `IsNull` method, which enables checking
+  if the value is null without type information and fully decoding underlying data'
+time: 2023-06-27T12:58:06.917152-04:00
+custom:
+  Issue: "305"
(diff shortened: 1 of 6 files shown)


bendbennett (approved):
LGTM 🚀

PR #308, opened 2023-06-30
tftypes: Reduce AttributePath and Value type compute and memory usage

### LINKED ISSUE #307: tftypes: Remove Extraneous Memory Allocations in ApplyTerraform5AttributePathStep

```
v0.17.0
```

Code which generally calls `tftypes.Walk()`/`tftypes.Transform()` and involves collection/object types.

### Terraform Configuration Files

Code with high numbers of elements/attributes.

```tf
resource "hashicups_order" "example2" {
  # items is a set of objects with additional attributes beyond quantity
  items = [
    for e in range(1, 1000) : {
      quantity = e
    }
  ]
}
```

### Expected Behavior

Reasonable performance of `tftypes.Walk()` and `tftypes.Transform()` calls, let's arbitrarily say less than 1 second for handling a set of 1000 object elements.

### Actual Behavior

Operations take many seconds with 1000 set of object elements, with a particular hot spot in `(tftypes.Value).ApplyTerraform5AttributePathStep()`. The code in that method is unnecessarily re-allocating an entire map/slice only to return the element at a map key or slice index. This causes many memory allocations on the heap and GC cycles.

Screenshot 2023-06-29 at 8 09 28 AM

Screenshot 2023-06-29 at 9 12 59 AM

### Steps to Reproduce

Should be able to unit test benchmark this, but the real-world reproduction was done via terraform-provider-hashicups-pf.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/775

bflad:
Closes #307

These optimizations were performed by adding benchmark tests against a set of 1000 "simple" objects and viewing the cpu/memory profiles. The original implementations were spending an immense amount of time dealing with memory allocations and garbage collection, so reducing memory allocations was the main target of these optimizations, which in turn, also reduced compute time.

The largest changes were accomplished by removing `(Value).Diff()` from the logic paths which were only testing equality. The new `(Value).deepEqual()` implementation started as a duplicate of that logic, removing all `ValueDiff` allocations. Then, the walking of `Value` needed a methodology to stop the walk immediately to prevent further allocations. The two changes in that regard were introducing a `stopWalkError` sentinel error type for which callback functions can signal that no remaining `Value` are necessary to traverse and updating the internal `walk()` implementation to include a new bool return to fully accomplish the intended behavior.

The other changes were smaller optimizations noticed from the profiling, such as avoiding the Go runtime immediately reallocating a larger slice with the `AttributePath` methods, avoiding memory allocations caused by `(Type).Is()` usage instead of using type switches, and directly referencing `List`, `Map`, `Object`, `Set`, and `Tuple` value storage instead of allocating an unneeded variable. This logic is heavily used both by this Go module and others, so even these minor optimizations have impact at scale.

`benchstat` differences between prior implementation and proposed changes:

```
goos: darwin
goarch: arm64
pkg: github.com/hashicorp/terraform-plugin-go/tftypes
                                             │   original   │              proposed               │
                                             │    sec/op    │   sec/op     vs base                │
Transform1000-10                               2886.1µ ± 0%   924.6µ ± 0%  -67.96% (p=0.000 n=10)
ValueApplyTerraform5AttributePathStep1000-10   3863.4µ ± 1%   628.9µ ± 0%  -83.72% (p=0.000 n=10)
geomean                                         3.339m        762.5µ       -77.16%

                                             │   original    │               proposed               │
                                             │     B/op      │     B/op      vs base                │
Transform1000-10                               3137.3Ki ± 0%   837.6Ki ± 0%  -73.30% (p=0.000 n=10)
ValueApplyTerraform5AttributePathStep1000-10   3887.1Ki ± 0%   389.3Ki ± 0%  -89.98% (p=0.000 n=10)
geomean                                         3.410Mi        571.0Ki       -83.65%

                                             │   original   │              proposed               │
                                             │  allocs/op   │  allocs/op   vs base                │
Transform1000-10                                61.07k ± 0%   14.02k ± 0%  -77.05% (p=0.000 n=10)
ValueApplyTerraform5AttributePathStep1000-10   123.07k ± 0%   17.64k ± 0%  -85.67% (p=0.000 n=10)
geomean                                         86.69k        15.72k       -81.86%
```

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230630-103655.yaml
+++ b/.changes/unreleased/BUG FIXES-20230630-103655.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'tftypes: Significantly reduced compute and memory usage of `Value` type
+  walking and transformation'
+time: 2023-06-30T10:36:55.119505-04:00
+custom:
+  Issue: "307"
--- a/.changes/unreleased/BUG FIXES-20230630-144701.yaml
+++ b/.changes/unreleased/BUG FIXES-20230630-144701.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'tftypes: Removed unnecessary memory allocations from `AttributePath` type
+  `Equal()`, `LastStep()`, and `WithoutLastStep()` methods'
+time: 2023-06-30T14:47:01.127283-04:00
+custom:
+  Issue: "307"
--- a/.changes/unreleased/BUG FIXES-20230630-144945.yaml
+++ b/.changes/unreleased/BUG FIXES-20230630-144945.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'tftypes: Removed unnecessary memory allocations from `NewValue()` function'
+time: 2023-06-30T14:49:45.828209-04:00
+custom:
+  Issue: "307"
--- a/.changes/unreleased/ENHANCEMENTS-20230630-143817.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20230630-143817.yaml
@@ -1,0 +1,6 @@
+kind: ENHANCEMENTS
+body: 'tftypes: Added `AttributePath` type `NextStep()` method, which returns the
+  next step in the path without first copying via `Steps()`'
+time: 2023-06-30T14:38:17.600608-04:00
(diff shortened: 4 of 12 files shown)


bendbennett (approved):
LGTM 🚀

PR #311, opened 2023-07-07
tfprotov5+tfprotov6: Update to protocol versions 5.4/6.4 with GetMetadata RPC and GetProviderSchemaOptional server capability

### LINKED ISSUE #310: Introduce Protocol Server Capability for Optional GetProviderSchema RPC

```
v0.18.0
```

### Use cases

Terraform core is working on optimizing their overall memory utilization by supporting the ability to cache provider schemas per-type, rather than per-instance. This should cause a statistically significant drop in maximum resident size in Terraform environments running configurations with multiple provider aliases for larger providers, such as `hashicorp/aws`. The challenge is that not all provider implementations support all RPC operations without the historically upfront `GetProviderSchema` RPC, such as older versions of terraform-plugin-framework, which used the `GetProviderSchema` RPC to cache supported resource types for other RPCs.

### Proposal

Update the plugin protocol to 5.4 and 6.4 with a new server capability to signal that the `GetProviderSchema` RPC is optional/skippable. Initial testing shows that terraform-plugin-sdk/v2@v2.27.0 and terraform-plugin-framework@v1.3.2 should be able to enable this capability once its exposed by this Go module. In terraform-plugin-mux, it should be able to front each existing RPC with a logic to call `GetProviderSchema` first if necessary for the underlying provider server.

### References

- https://github.com/hashicorp/terraform/pull/33486
- https://github.com/hashicorp/terraform/pull/33482
- https://github.com/hashicorp/terraform-plugin-mux/issues/185
- https://github.com/hashicorp/terraform-plugin-framework/issues/828
- https://github.com/hashicorp/terraform-plugin-sdk/issues/1234

bflad:
Closes #310

The CI failures for terraform-provider-corner are expected in this case. Protocol upgrades that impose new RPCs will either require:

- The `tfprotov5.ProviderServer`/`tfprotov6.ProviderServer` interface to require a new method
- Or, new "optional" interfaces be implemented (let's make up `ProviderServerWithGetMetadata` in this case)

terraform-plugin-go is a low level abstraction which is designed to _directly implement_ the protocol rather than introduce its own abstractions. Most provider developers will directly interface with higher level Go modules, such as terraform-plugin-sdk and terraform-plugin-framework. Except for advanced provider development using this Go module directly, this type of low level "breaking" change will be hidden by also upgrading those Go modules at the same time:

- https://github.com/hashicorp/terraform-plugin-sdk/pull/1235
- https://github.com/hashicorp/terraform-plugin-framework/pull/829
- https://github.com/hashicorp/terraform-plugin-mux/pull/186

Therefore the change is implemented on the existing interface. terraform-provider-corner happens to also implement its own terraform-plugin-go based provider server code, which is affected by this change. Upgrading the various dependencies and implementing the new method makes terraform-provider-corner testing successful, which can be seen with https://github.com/hashicorp/terraform-provider-corner/pull/174. The CI of this Go module and terraform-provider-corner implementation details just happens to be a chicken-and-egg problem which tends to only come up in this particular type of functionality change. 🐔 🥚


austinvalle (approved):
LGTM 🚀

bendbennett (approved):
LGTM 🚀

bflad:
`GetProviderSchemaOptional` works as expected with upstream https://github.com/hashicorp/terraform/pull/33807 and downstream https://github.com/hashicorp/terraform-plugin-framework/pull/829 / https://github.com/hashicorp/terraform-plugin-sdk/pull/1235 / https://github.com/hashicorp/terraform-plugin-mux/pull/186. Merging!

PR #354, opened 2023-12-14
tfprotov5+tfprotov6: Require FunctionServer in ProviderServer

### LINKED ISSUE #353: tfprotov5+tfprotov6: Add FunctionServer to ProviderServer interface

```
v0.20.0 (not yet released)
```

### Description

The upcoming v0.20.0 release will contain the `tfprotov5` and `tfprotov6` package `FunctionServer` interface, which contains the two RPC function handlers for provider defined functions. Rather than immediately requiring all downstream SDK implementations to implement this functionality immediately by requiring `FunctionServer` as part of `ProviderServer`, it seems ideal if we can stagger the releases between introducing and supporting the new `FunctionServer` bits then having a followup release that makes the implementation required with the "breaking change" to match the intended design of this Go module to always require all provider implementations to fully implement the protocol.

For provider developers, this type of terraform-plugin-go change previously surfaced as confusing `missing method XXX` when updating only some dependencies of terraform-plugin-framework, terraform-plugin-mux, terraform-plugin-sdk, or terraform-plugin-testing where terraform-plugin-go required the additional `ProviderServer` bits, but with the newer setup there can at least be better chance that particular update issue won't occur for them.

### Proposal

In the first release, temporarily check for `FunctionServer` interface implementation in `tfprotov5/tf5server` and `tfprotov6/tf6server` package `GetFunctions` and `CallFunction` implementations. If found, pass request downstream. If not found, raise an error diagnostic.

In the second release (part of this issue), add `FunctionServer` interface in `ProviderServer` implementation. Remove temporary `FunctionServer` interface checks and associated error diagnostic.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/861
- https://github.com/hashicorp/terraform-plugin-testing/issues/185

bflad:
Closes #353

This breaking change aligns the `tfprotov5` and `tfprotov6` packages with the intended design of this Go module, which is to fully implement the Terraform Plugin Protocol. The `FunctionServer` interface was temporarily optional in v0.20.0 so downstream SDKs could implement and release support first with a lesser likelihood that developers would run into compilation errors when only upgrading a single downstream Go module.

Provider developers should upgrade this Go module only after ensuring their provider Go module is using at least:

- github.com/hashicorp/terraform-plugin-framework@v1.5.0
- github.com/hashicorp/terraform-plugin-mux@v0.13.0
- github.com/hashicorp/terraform-plugin-sdk/v2@v2.31.0

### DIFF
--- a/.changes/unreleased/BREAKING CHANGES-20231214-105454.yaml
+++ b/.changes/unreleased/BREAKING CHANGES-20231214-105454.yaml
@@ -1,0 +1,8 @@
+kind: BREAKING CHANGES
+body: 'tfprotov5+tfprotov6: The `ProviderServer` interface now requires the `FunctionServer`
+  interface. Before upgrading this Go module, ensure you are depending on at least
+  terraform-plugin-framework v1.5.0, terraform-plugin-mux v0.13.0, and
+  terraform-plugin-sdk/v2 v2.31.0.'
+time: 2023-12-14T10:54:54.463017-05:00
+custom:
+  Issue: "353"
--- a/tfprotov5/provider.go
+++ b/tfprotov5/provider.go
@@ -53,10 +53,7 @@ type ProviderServer interface {
 	// are a handy interface for defining what a function is to
 	// terraform-plugin-go, so they are their own interface that is composed
 	// into ProviderServer.
-	//
-	// This will be required in an upcoming release.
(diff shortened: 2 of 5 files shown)


bflad:
Replaced with #388 👍

Issue #363, opened 2024-01-12
tfprotov5+tfprotov6: Move MoveResourceState method into ResourceServer


bflad:
### terraform-plugin-go version

```
v0.21.0 (not yet released)
```

### Description

The upcoming v0.21.0 release will contain the `tfprotov5` and `tfprotov6` package `ResourceServerWithMoveResourceState` temporary interface, which contains the `MoveResourceState` RPC function handler. Rather than immediately requiring all downstream SDK implementations to implement this functionality immediately by requiring `MoveResourceState` as part of `ResourceServer`, it seems ideal if we can stagger the releases between introducing and supporting the new `MoveResourceState` bits then having a followup release that makes the implementation required with the "breaking change" to match the intended design of this Go module to always require all provider implementations to fully implement the protocol.

For provider developers, this type of terraform-plugin-go change previously surfaced as confusing `missing method XXX` when updating only some dependencies of terraform-plugin-framework, terraform-plugin-mux, terraform-plugin-sdk, or terraform-plugin-testing where terraform-plugin-go required the additional `ResourceServer` bits, but with the newer setup there can at least be better chance that particular update issue won't occur for them and changelog can reference the minimum SDK versions.

### Proposal

Move the `MoveResourceState` method from `ResourceServerWithMoveResourceState` to `ResourceServer`. Remove temporary `ResourceServerWithMoveResourceState` interface, logic referencing it, and associated error diagnostics when it was missing in the server implementations.

### References

- #362

bflad:
Per https://github.com/hashicorp/terraform-plugin-go/pull/388#issuecomment-1991144126, this issue is going to be changed into just copying the `MoveResourceState` method into the `ResourceServer` interface and a separate issue created for the actual `ResourceServerWithMoveResourceState` interface removal.

PR #364, opened 2024-01-12
tfprotov5+tfprotov6: Initial MoveResourceState support

### LINKED ISSUE #362: tfprotov5+tfprotov6: Implement MoveResourceState RPC

```
v0.20.0
```

### Use cases

Protocol versions 5.5/6.5, as part of Terraform 1.8, will also be introducing the `MoveResourceState` RPC, which is intended to enable provider developers to support state move across resource types. In practice, this enables use cases such as renaming resource types or migrating between known-compatible providers with differing resource type names.

### Proposal

The `MoveResourceState` RPC handling is built on top of already existing data handling types, so this implementation is relatively straightforward in the `tfprotov5` and `tfprotov6` packages:

* Implement the new `MoveResourceState` server capability so the providers can announce they support this new RPC.
* Implement a temporary `ResourceServerWithMoveResourceState` interface that implements new `MoveResourceState` method, along with its request and response types.
* Update `tf5server`/`tf6server` implementations for the new RPC and check for the temporary interface.

Another issue will be created to perform the breaking change of moving the `MoveResourceState` method into the existing `ResourceServer` interface. This will allow a grace period for all officially supported SDKs to implement the new method without the breaking change, then the breaking change can be announced with the required versions of each SDK all at once when they upgrade this Go module to the version that requires the implementation.

### References

- https://github.com/hashicorp/terraform/pull/34480

bflad:
Closes #362
Reference: https://github.com/hashicorp/terraform-plugin-go/issues/363

The next version of the plugin protocol (5.5/6.5) includes support for moving resource state. This change introduces the initial implementation of that support including:

- Updated Protocol Buffers definitions
- Re-generated Protocol Buffers Go code
- Initial implementations of `tfprotov5` and `tfprotov6` package abstractions and wiring between those abstractions and the Protocol Buffers generated Go code
- Initial implementations of `tfprotov5/tf5server` and `tfprotov6/tf6server` for the new `MoveResourceState` RPC

This temporarily will not require `ProviderServer` implementations to include `MoveResourceState` implementation, however that change will occur in a subsequent release.


bendbennett (approved):
LGTM 🚀

austinvalle (approved):
:shipit:

Issue #365, opened 2024-01-12
tfprotov5+tfprotov6: Cleanup and unit test fromproto/toproto packages


bflad:
### terraform-plugin-go version

```
v0.20.0
```

### Description

The `tfprotov(5|6)/internal/(from|to)proto` packages are there for converting to/from the provider-facing types and the Protocol Buffers generated code types. There are a few little issues in those packages that would be nice to handle for future travelers:

* `fromproto` containing response types and `ServerCapabilities`, which are extraneous
* `toproto` containing request types, which are extraneous
* `DynamicValue` conversions not handling `nil` case in the function
* Lack of proper unit testing to ensure type mapping coverage

PR #369, opened 2024-01-23
tfprotov5+tfprotov6: Remove unnecessary AttributePath_Step error return

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-go/issues/365

It is not possible for an unexpected attribute path step implementation to reach the protocol conversion logic unless there is a protocol change, which the logic should be updated anyways at the same time and unit tested. This simplifies many of the gRPC server handling methods to not need to handle the currently impossible-to-have errors.

### DIFF
--- a/tfprotov5/internal/toproto/attribute_path.go
+++ b/tfprotov5/internal/toproto/attribute_path.go
@@ -4,95 +4,89 @@
 package toproto
 
 import (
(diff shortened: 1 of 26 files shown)


austinvalle (approved):
Clean! :shipit: 🧹 🧹

PR #372, opened 2024-01-25
tfprotov5+tfprotov6: Include private state in MoveResourceState

bflad:
Reference: https://github.com/hashicorp/terraform/pull/34575

To match upstream changes. No changelog is necessary as this is unreleased.


bendbennett (approved):
LGTM 🚀

PR #381, opened 2024-02-21
all: Update Go Module to Go 1.21

bflad:
Reference: https://github.com/hashicorp/terraform-providers-devex-internal/issues/175


bendbennett (approved):
LGTM 🚀

PR #384, opened 2024-02-29
Adding go toolchain directive

bendbennett:
Reference: https://github.com/hashicorp/terraform-providers-devex-internal/issues/175

- Reinstating the `go toolchain` directive in `go.mod`.


austinvalle (approved):
lgtm 🚀 , ref: https://github.com/hashicorp/terraform-providers-devex-internal/issues/175

PR #395, opened 2024-04-16
workflows: Remove wildcard suffix from Terraform workflow call

austinvalle:
Can be merged after -> https://github.com/hashicorp/terraform-devex-repos/pull/69


SBGoods (approved):

austinvalle:
Variable updated + re-ran the workflows and we're all green 💚 🚀

PR #396, opened 2024-04-17
don't marshal integer values as msgpack floats

### LINKED ISSUE #391: Coordinate switch in message pack encoding for large integers with Terraform core

```
v0.22.1
```

An issue describing that [Planned value does not match config value for number](https://github.com/hashicorp/terraform/issues/34866#top) on Terraform core has resulted in a fix in cty in v1.14.4 (refer to https://github.com/hashicorp/terraform/issues/34866#issuecomment-2010895435). The fix was to change the MessagePack encoder to use a different encoding for large integers. This should fix the core-to-provider direction in terraform-plugin-go, but equivalent changes will need to implemented in this repository for the provider-to-core direction.

The changes to Terraform core and terraform-plugin-go should be coordinated to minimise any time during which encodings differ for Terraform core and terraform-plugin-go.

### References

- hashicorp/terraform#34866
- https://github.com/zclconf/go-cty/commit/f41ae52fdfa8a9590aa00c3ab3ff13cef4dd872f

jbardin:
Round-tripping a large integer through a float64, even if the binary representation is exact, causes us to end up with a rounded string representation after decoding, because the decoded number has 52-bit precision instead of the 512 we use when dealing with string representations of large integers.

The test fixture was changed slightly to compare only the final value string representations, which more closely corresponds to the configuration and state values. The comparison of the `Value` directly will contain differences in the underlying `math/big` value.

This mirrors the upstream change in go-cty [here](https://github.com/zclconf/go-cty/commit/f41ae52fdfa8a9590aa00c3ab3ff13cef4dd872f), which will be incorporated into the next version of Terraform, ensuring that the round-trip encoding of values is the same in both directions.

Fixes #391

### DIFF
--- a/.changes/unreleased/BUG FIXES-20240418-121811.yaml
+++ b/.changes/unreleased/BUG FIXES-20240418-121811.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'tftypes: Large integers are always encoded as msgpack integers rather than
+  float values to ensure the decoded value will not be rounded to 52-bit precision'
+time: 2024-04-18T12:18:11.880247-04:00
+custom:
+  Issue: "396"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.22.2 (Unreleased)
+
+BUG FIXES:
+
+* tftypes: Large integers are always encoded as msgpack integers rather than float values to ensure the decoded value will not be rounded to 52-bit precision ([#396](https://github.com/hashicorp/terraform-plugin-go/pull/396))
+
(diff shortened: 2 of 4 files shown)


austinvalle (approved):
LGTM, I'll push a changelog to your branch before merging 👍🏻 

> Note: You'll see me open another PR temporarily with these exact changes because our workflows are currently busted when a PR comes from a fork and don't run our CI tests 🙃

austinvalle:
All workflows are green over in https://github.com/hashicorp/terraform-plugin-go/actions/runs/8741201175?pr=397 🟢

Thanks for the fix @jbardin!

PR #403, opened 2024-05-01
tfprotov5+tfprotov6: Add deferred action support to related RPCs

austinvalle:
Refs: https://github.com/hashicorp/terraform/pull/34880, https://github.com/hashicorp/terraform/pull/35063

This PR introduces the `ClientCapabilities/deferral_allowed` fields to the relevant request structs and `Deferred/Reason` to the relevant response structs. The RPCs in-scope for supporting deferred actions are:
- `ReadDataSource`
- `ReadResource`
- `PlanResourceChange`
- `ImportResourceState`

The `ConfigureProvider` RPC also includes `ClientCapabilities/deferral_allowed` to assist provider logic that needs to automatically setup deferral responses for the `PROVIDER_CONFIG_UNKNOWN` reason.


tfprotov5/client_capabilities.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package tfprotov5
+
+// ClientCapabilities allows Terraform to publish information regarding optionally supported
+// protocol features, such as forward-compatible Terraform behavior changes.
+type ClientCapabilities struct {
+	// DeferralAllowed signals that the Terraform client is able to
+	// handle deferred responses from the provider.
+	DeferralAllowed bool
```
bflad, line 11: Nit: For future travelers, I wonder if it would be worth calling out which RPCs this applies to, otherwise I worry about this struct having a lot of fields that only sometimes apply to whatever RPC its attached to.

Another option would be creating RPC-specific `ClientCapabilities` struct types (e.g. `ConfigureProviderClientCapabilities` 😅 ) to protect the exported API from changing in a future where the protocol potentially introduces RPC-specific capabilities. We had opted for the protocol to share the message for now and changing that in the protocol is safe to split up as it only affects the generated code, but exported Go package code not so much.

bflad (commented):
So far this looks good to me! One potential consideration to document and/or futureproof the exported API for client capabilities. 👍

bflad:
Oh and for your explicit open questions -- these both seem like things that can be added now or later, if we want. The logging might be nice and easy to get in right now though. The diagnostic wouldn't be hard to implement, but it'll be one of the first times introducing automatic RPC-level diagnostics that cannot be overridden by provider developers or higher level SDKs. Not sure if there are any concerns about this codebase owning an unavoidable error and its messaging (probably not?) but I have to ask.

tfprotov5/client_capabilities.go
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tfprotov5
+
+// ClientCapabilities allows Terraform to publish information regarding optionally supported
+// protocol features, such as forward-compatible Terraform behavior changes.
+type ClientCapabilities struct {
+	// DeferralAllowed signals that the Terraform client is able to
```
austinvalle, line 9: **Reminder from offline convo:** Update these pkg docs to reflect that the client capability is related to the specific request and not necessarily the client

tfprotov5/client_capabilities.go
austinvalle: Yeah this is a good point, it feels very likely this would happen, especially with the design that the capabilities are always related to specific requests and not the client as a whole.

I'll duplicate all of these to structs as `{RPCName}ClientCapabilities` to protect us from that but keep the actual property name as `ClientCapabilities`.

cc @SBGoods -> just a heads up that there will be an upcoming change once I adjust this

austinvalle:
@bflad Cool, I'll implement the logging since it's straightforward.

As for the diagnostic, I'm 50/50 on adding the error check... On one hand, it's an implicit expectation of the protocol that Terraform core has which is not immediately obvious without documentation. On the other, there are other implicit expectations around combinations of data.

I'm leaning towards adding the error check, but I'm going to verify with core to ensure that applying this error handling at the protocol level is appropriate. I feel like if they have an error check on their side for this, we should as well.

Both Framework and SDKv2 will have their own error checks for this scenario, so regardless, it will be handled by the higher-level SDK with it's messaging.

tfprotov5/client_capabilities.go
austinvalle: Split all of the structs up -> 3a4fbc8 ⚡

tfprotov5/client_capabilities.go
austinvalle: Fixed pkg comments in 3a4fbc8 ⚡

austinvalle:
@bflad I've made all of the changes we discussed so I think it's ready for the PR to be officially reviewed 😆 

- Added logging in a58e305 ⚡ 
- Updated the protocol version variables. Also added two additional tests that do a simple compare on the proto file comments to find a match. If we find it to be too cumbersome to maintain we can always remove it later. Example of test failures before updating the variable:
```bash
--- FAIL: Test_EnsureVersionConstantMatchesProtoFile (0.00s) proto_version_test.go:47: protocol version Go variable is different from proto file - expected: 5.6, got: 5.4
proto_version_test.go:48: MAINTAINER NOTE: Update tf5server.protocolVersionMajor and tf5server.protocolVersionMinor to match the proto file.

--- FAIL: Test_EnsureVersionConstantMatchesProtoFile (3.59s) proto_version_test.go:47: protocol version Go variable is different from proto file - expected: 6.6, got: 6.4
proto_version_test.go:48: MAINTAINER NOTE: Update tf6server.protocolVersionMajor and tf6server.protocolVersionMinor to match the proto file.
```

- Added a diagnostic for invalid deferred responses. I don't see any tests for the `tf{5/6}server` packages, should I add some or is it fine to just test this in `terraform-provider-corner` with a `terraform-plugin-go` provider?

tfprotov5/internal/tf5serverlogging/client_capabilities.go
```diff
@@ -0,0 +1,76 @@
...
+	responseFields := map[string]interface{}{
+		logging.KeyClientCapabilityDeferralAllowed: false,
+	}
+
+	if capabilities != nil {
+		responseFields[logging.KeyClientCapabilityDeferralAllowed] = capabilities.DeferralAllowed
+	}
+
+	logging.ProtocolTrace(ctx, "Announced client capabilities", responseFields)
+}
```
bflad, line 24: I think this setup is reasonable to show the TRACE log even with older versions of Terraform that wouldn't have this client capability -- otherwise developers/maintainers will need to pay attention to the omission of the log entry instead. Another way to do this logging would be to key a different "No announced client capabilities" log message when `capabilities != nil`, but the current setup is also fine in my book. 👍

tfprotov5/tf5server/server.go
```diff
@@ -692,6 +694,17 @@ func (s *server) ReadDataSource(ctx context.Context, protoReq *tfplugin5.ReadDat
...
+	tf5serverlogging.Deferred(ctx, resp.Deferred)
+
+	if resp.Deferred != nil && (req.ClientCapabilities == nil || !req.ClientCapabilities.DeferralAllowed) {
+		resp.Diagnostics = append(resp.Diagnostics, &tfprotov5.Diagnostic{
+			Severity: tfprotov5.DiagnosticSeverityError,
+			Summary:  "Invalid Deferred Response",
+			Detail: "Provider returned a deferred response but the Terraform request did not indicate support for deferred actions." +
+				"This is an issue with the provider and should be reported to the provider developers.\n\n" +
+				fmt.Sprintf("Deferred reason - %q", resp.Deferred.Reason.String()),
+		})
```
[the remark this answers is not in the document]
bflad: Looks good to me!

tfprotov5/internal/tf5serverlogging/client_capabilities.go
austinvalle: Does f029631 match what you were thinking?

bflad: f02963174f74fb70d8d39c47b95f94adeade54e0 does indeed match! Thank you.

SBGoods (approved):
LGTM, Great job!

PR #408, opened 2024-05-10
tfprotov5+tfprotov6: Remove temporary ResourceServerWithMoveResourceState interface

### LINKED ISSUE #389: tfprotov5+tfprotov6: Remove temporary ResourceServerWithMoveResourceState interface

```
v0.23.0
```

### Description

The `ResourceServerWithMoveResourceState` temporary interface was introduced as a way to onboard downstream SDKs onto the new `MoveResourceState` RPC before it was required. This enabled a period where terraform-plugin-go could support new functionality while not causing compilation errors if downstream Go modules happened to be referencing differing SDK versions that may or may not all support the new functionality immediately.

The challenge with this setup is that terraform-plugin-mux is low level enough to need to reference even the temporary interface in order to work correct. This means that while `ResourceServer` can require the new method, the temporary interface needs one additional release before it can removed "safely".

### Proposal

In `tfprotov5` and `tfprotov6` remove the `ResourceServerWithMoveResourceState` interface type. In the changelog, note that consumers should require terraform-plugin-mux@v0.16.0 (or whatever is correct at the time) dependency to prevent compilation errors.

### References

- https://github.com/hashicorp/terraform-plugin-go/pull/388

bflad:
Closes #389

The `ResourceServerWithMoveResourceState` temporary interface was introduced as a way to onboard downstream SDKs onto the new `MoveResourceState` RPC before it was required. This enabled a period where terraform-plugin-go could support new functionality while not causing compilation errors if downstream Go modules happened to be referencing differing SDK versions that may or may not all support the new functionality immediately. Now that those downstream Go modules are referencing `ResourceServer` directly again, the temporary interface can be removed to prevent bloating this Go module.

### DIFF
--- a/.changes/unreleased/BREAKING CHANGES-20240510-105824.yaml
+++ b/.changes/unreleased/BREAKING CHANGES-20240510-105824.yaml
@@ -1,0 +1,6 @@
+kind: BREAKING CHANGES
+body: 'tfprotov5+tfprotov6: Removed temporary `ResourceServerWithMoveResourceState`
+  interface type. Use `ResourceServer` instead.'
+time: 2024-05-10T10:58:24.383876-04:00
+custom:
+  Issue: "408"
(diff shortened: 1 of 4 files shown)


austinvalle (approved):
🎉

PR #462, opened 2025-01-17
tfprotov5+tfprotov6: Write-only Attribute Implementation

SBGoods:
The next version of the plugin protocol (5.8/6.8) introduces a new `write_only` property to the `Schema.Attribute` message to specify a write-only attribute. Write-only attributes are attributes which do not have their values persisted to Terraform state. `write_only` is only applicable for managed resource attributes.


tfprotov5/client_capabilities.go
```diff
@@ -3,6 +3,15 @@
 
 package tfprotov5
 
+// ValidateResourceTypeConfigClientCapabilities allows Terraform to publish information
```
austinvalle, line 6: My monthly reminder that this RPC is named different in protocol v5 😆

tfprotov5/client_capabilities.go
austinvalle, line 1: Not specific feedback to this work, more of an observation, but I see that Terraform core is sending these client capabilities in every RPC request regardless of whether we use them. So in theory we could add, say, the `WriteOnlyAttributesAllowed` capability to `PlanResourceChange` with an SDK only change.

Not saying we should, but thought it was interesting.

tfprotov6/schema.go
```diff
@@ -253,6 +253,12 @@ type SchemaAttribute struct {
 	// experiences. Providers should set it when deprecating attributes in
 	// preparation for these tools.
 	Deprecated bool
+
+	// WriteOnly, when set to true, indicates that the attribute value will
+	// be provided via configuration and must be omitted from state. WriteOnly
```
austinvalle, line 258:
```suggestion
	// be provided via configuration and must be omitted from plan and state response objects. WriteOnly
```

tfprotov5/schema.go
```diff
@@ -221,6 +221,12 @@ type SchemaAttribute struct {
 	// experiences. Providers should set it when deprecating attributes in
 	// preparation for these tools.
 	Deprecated bool
+
+	// WriteOnly, when set to true, indicates that the attribute value will
+	// be provided via configuration and must be omitted from state. WriteOnly
```
austinvalle, line 226:
```suggestion
	// be provided via configuration and must be omitted from plan and state response objects. WriteOnly
```

austinvalle (approved):
LGTM, sorry I have one suggestion in the doc string just to be super explicit about the field

austinvalle (approved):

PR #474, opened 2025-02-07
Added testing for tfprotov5

rainkwan:
Added testing for `toproto/resource_identity_data`, `toproto/resource_identity_schema`, (Renamed as per suggested in last PR) and `toproto/resource`. Also started to adding to `logging/keys` and `logging/context` so that we can next start to update `logging/server`

PR #476, opened 2025-02-10
tfprotov6: Resource Identity implementation

ansgarm:
shamelessly copied from protocol v5 as merged via #474

differences to the other PR:
* this PR uses `v6` where the other used `v5`
* the server.go for `v6` was in a sub directory

This PR also adds two changie entries for this change and the two related ones that already landed

### DIFF
--- a/.changes/unreleased/NOTES-20250210-175414.yaml
+++ b/.changes/unreleased/NOTES-20250210-175414.yaml
@@ -1,0 +1,5 @@
+kind: NOTES
+body: 'tfprotov5+tfprotov6: An upcoming release will require the `ProviderServerWithResourceIdentity` implementation as part of `ProviderServer`.'
+time: 2025-02-10T17:54:14.054598+01:00
+custom:
+    Issue: "476"
--- a/.changes/unreleased/NOTES-20250210-175414.yaml
+++ b/.changes/unreleased/NOTES-20250210-175414.yaml
@@ -1,5 +1,5 @@
 kind: NOTES
-body: 'tfprotov5+tfprotov6: An upcoming release will require the `ProviderServerWithResourceIdentity` implementation as part of `ProviderServer`.'
+body: 'tfprotov5+tfprotov6: An upcoming release will require the `GetResourceIdentitySchemas` and `UpgradeResourceIdentity` implementations as part of `ProviderServer`.'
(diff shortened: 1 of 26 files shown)


.changes/unreleased/NOTES-20250210-175414.yaml
```diff
@@ -0,0 +1,5 @@
+kind: NOTES
+body: 'tfprotov5+tfprotov6: An upcoming release will require the `ProviderServerWithResourceIdentity` implementation as part of `ProviderServer`.'
```
austinvalle, line 2: nit: maybe a little more specific
```suggestion
body: 'tfprotov5+tfprotov6: An upcoming release will require the `GetResourceIdentitySchemas` and `UpgradeResourceIdentity` implementations as part of `ProviderServer`.'
```

tfprotov6/provider.go
```diff
@@ -147,6 +170,26 @@ type GetProviderSchemaResponse struct {
...
 }
 
+// GetResourceIdentitySchemasRequest represents a Terraform RPC request for the
+// provider's resource identity schemas.
+type GetResourceIdentitySchemasRequest struct{}
+
+// GetResourceIdentitySchemasResponse represents a Terraform RPC response containing
+// the provider's resource identity schemas.
+type GetResourceIdentitySchemasResponse struct {
+	// ResourceSchemas is a map of resource names to the schema for the
```
austinvalle, line 180: I missed this in the other PR 😄 , I believe v5 has the same mistype
```suggestion
	// IdentitySchemas is a map of resource names to the schema for the
```

austinvalle (approved):
Two comments! LGTM 👍🏻

tfprotov6/provider.go
ansgarm: Good catch! Yeah, v5 had the same one – I just added an commit to this PR to fix that while we're at it

austinvalle (approved):
LGTM, the CI errors are all related to a new golangci-lint version (it gets shadow bumped by the github action 🙁 ), so you can just go ahead and merge 🚀 

Opened a PR to fix the lint errors: #477

austinvalle:
Rebased, and now you have some actual lint errors in this PR to fix 🤣 , apologies!

austinvalle (approved):
🚀

PR #477, opened 2025-02-12
chore: Update golangci-lint linters and apply fixes

austinvalle:
> Follow-up to #476 where we found the linter deprecations

If you're not familiar what the fix/original behavior was (or are like me and completely forgot since we're always 1 go version behind 😆): [Fixing For Loops in Go 1.22](https://go.dev/blog/loopvar-preview)


ansgarm (approved):
image

PR #484, opened 2025-03-07
feat: Add `WithManagedDebugEnvFilePath()` option that allows writing `TF_REATTACH_PROVIDERS` to an environment file

ansgarm:
This PR introduces the option to write the `TF_REATTACH_PROVIDERS` environment variable (and its value) into a file that can be used by downstream tooling.
For example, it can be used in a VSCode launch configuration for debugging Terraform, which allows to debug both provider and Terraform at the same time without the constant need to change `TF_REATTACH_PROVIDERS` in Terraform's debug configuration:

image


austinvalle (approved):
Nice addition! LGTM 🚀

PR #487, opened 2025-03-11
Updated protocol to reflect core changes to protocol move request

rainkwan:
Updated protov5 and v6 to reflect changes to raw_identity introduced by [core updates to protocol](https://github.com/hashicorp/terraform/pull/36662/files#diff-275921a6ad82cf3cf2e7cee07076a2bb4f168848f081b2b7f76c2c7966d5c6cbR479 )


tfprotov5/resource.go
```diff
@@ -145,7 +145,7 @@ type UpgradeResourceIdentityRequest struct {
 	// RawIdentity is the identity state as Terraform sees it right now. See the
 	// documentation for `RawIdentity` for information on how to work with the
 	// data it contains.
-	RawIdentity *RawIdentity
+	RawState *RawState
```
austinvalle, line 148: We should keep this matching the protocol, which is still named `raw_identity`

```suggestion
	RawIdentity *RawState
```

tfprotov6/resource.go
```diff
@@ -142,7 +142,7 @@ type UpgradeResourceIdentityRequest struct {
 	// RawIdentity is the identity state as Terraform sees it right now. See the
 	// documentation for `RawIdentity` for information on how to work with the
 	// data it contains.
-	RawIdentity *RawIdentity
+	RawState *RawState
```
austinvalle, line 145: Similar comment to the protov5 one

```suggestion
	RawIdentity *RawState
```

tfprotov5/internal/fromproto/raw_identity.go
```diff
@@ -1,20 +0,0 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
-
-package fromproto
-
-import (
-	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
-)
-
-func RawIdentity(in []byte) *tfprotov5.RawIdentity {
```
austinvalle, line 10: Since there is no longer a concept of a `tfprotov5.RawIdentity`, can we delete the struct/tests as well?

At `tfprotov{5/6}/raw_identity.go`

tfprotov6/resource.go
```diff
@@ -606,7 +606,10 @@ type MoveResourceStateRequest struct {
 	TargetTypeName string
 
 	// SourceIdentity is the identity of the source resource.
-	SourceIdentity *ResourceIdentityData
+	SourceIdentity *RawState
```
austinvalle, line 609: Similar comment to protov5

```suggestion
	//
	// Only the underlying JSON field is populated.
	SourceIdentity *RawState
```

tfprotov5/resource.go
```diff
@@ -609,7 +609,10 @@ type MoveResourceStateRequest struct {
 	TargetTypeName string
 
 	// SourceIdentity is the identity of the source resource.
-	SourceIdentity *ResourceIdentityData
+	SourceIdentity *RawState
```
austinvalle, line 612: Probably worth clarifying now, similar to `SourceState`
```suggestion
	//
	// Only the underlying JSON field is populated.
	SourceIdentity *RawState
```

austinvalle (commented):
Changes look good! Two minor comments to consider

austinvalle:
Also the CI might be failing because of a tool version mismatch?

Try running
```bash
make tools # install the latest tools
make protoc # generate the bindings again
```

SBGoods (approved):

austinvalle (approved):
I think we should still delete the files mentioned [here](https://github.com/hashicorp/terraform-plugin-go/pull/487#discussion_r1989787524), but that can come in a follow-up PR as well!

🚀

tfprotov5/internal/fromproto/raw_identity.go
ansgarm: here's a PR with that: https://github.com/hashicorp/terraform-plugin-go/pull/489

PR #514, opened 2025-05-20
rk/list-rpcs

rainkwan:

PR #517, opened 2025-05-21
tf5server: add ListResource RPC

bbasata:
## Related Issue

Fixes # 

## Description

In plain English, describe your approach to addressing the issue linked above. For example, if you made a particular design decision, let us know why you chose this path instead of another solution.

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### DIFF
--- a/tfprotov5/internal/fromproto/list_resource.go
+++ b/tfprotov5/internal/fromproto/list_resource.go
@@ -18,3 +18,15 @@ func ValidateListResourceConfigRequest(in *tfplugin5.ValidateListResourceConfig_
 		Config:   DynamicValue(in.Config),
 	}
 }
+
+func ListResourceRequest(in *tfplugin5.ListResource_Request) *tfprotov5.ListResourceRequest {
+	if in == nil {
+		return nil
+	}
+
+	return &tfprotov5.ListResourceRequest{
+		TypeName:              in.TypeName,
+		Config:                DynamicValue(in.Config),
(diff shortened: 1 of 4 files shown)


cmd/terraform-provider-primes/README.md
```diff
@@ -0,0 +1,71 @@
...
+            "type": "Im51bWJlciI=",
+            "description": "n"
+          }
+        ]
+      }
+    }
+  }
+}
+
+$ grpcurl -plaintext $(cmd/terraform-provider-primes/reattach.sh) tfplugin5.Provider.ListResource
```
bbasata, line 45: cc: @dbanck

bbasata:
Checks fail in the same way they do in #514. @rainkwan to 💭 with @austinvalle on whether a temporary interface is the way to go here or not.

austinvalle:
> Checks fail in the same way they do in #514. @rainkwan to 💭 with @austinvalle on whether a temporary interface is the way to go here or not.

Yeah temporary interfaces are good to use here to avoid the compilation errors. I'll write up a small blurb in this repo, but the general thing we need to avoid are temporary interfaces on global RPCs that Terraform uses the unimplemented GRPC code to determine if a provider supports the RPC.

That kind of RPC rarely pops up, and in the case of list, these RPCs aren't going to be reachable by today's workflows so  temporary is good 👍🏻

PR #523, opened 2025-05-28
actions interfaces

ansgarm:
- **tfprotov5: implement actions RPC calls and type converters**


tfprotov5/internal/streamproto/action.go
```diff
@@ -0,0 +1,32 @@
...
+	"google.golang.org/grpc"
+)
+
+func InvokeActionStreamingServer(in grpc.ServerStreamingServer[tfplugin5.InvokeAction_Event]) tfprotov5.InvokeActionStreamingServer {
+	return InvokeActionStreamingServerImpl{in}
+}
+
+var _ tfprotov5.InvokeActionStreamingServer = InvokeActionStreamingServerImpl{}
+
+type InvokeActionStreamingServerImpl struct {
```
ansgarm, line 21: unsure if `Impl` smells too much like Java 😆

PR #525, opened 2025-06-03
Add `ListResource` RPC

bbasata:
## Description

This change implements the Plugin Protocol v5 and v6 `ListResource` RPC. It includes an executable example in `examples/terraform-provider-primes`.

## Rollback Plan

None.


examples/terraform-provider-primes/provider.go
```diff
@@ -0,0 +1,144 @@
...
+			if err != nil {
+				panic(err)
+			}
+
+			protoEv := tfprotov5.ListResourceResult{
+				DisplayName: displayName,
+				Resource:    &resourceObject,
+			}
+			if !push(protoEv) {
+				fmt.Println("let's stop here")
```
ansgarm, line 124: curious: why do we stop? Is this the case where the caller effectively cancels?

tfprotov5/internal/fromproto/list_resource.go
```diff
@@ -8,6 +8,18 @@ import (
...
 
+func ListResourceRequest(in *tfplugin5.ListResource_Request) *tfprotov5.ListResourceRequest {
+	if in == nil {
+		return nil
+	}
+
+	return &tfprotov5.ListResourceRequest{
+		TypeName:        in.TypeName,
+		Config:          DynamicValue(in.Config),
+		IncludeResource: in.IncludeResourceObject,
```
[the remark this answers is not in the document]
ansgarm: (I might be completely wrong though, and this is a nit of course)

tfprotov5/internal/toproto/list_resource.go
```diff
@@ -18,6 +18,15 @@ func GetMetadata_ListResourceMetadata(in *tfprotov5.ListResourceMetadata) *tfplu
 	}
 }
 
+func ListResource_ListResourceEvent(in *tfprotov5.ListResourceResult) *tfplugin5.ListResource_Event {
+	return &tfplugin5.ListResource_Event{
+		DisplayName:    in.DisplayName,
+		ResourceObject: DynamicValue(in.Resource),
+		Identity:       ResourceIdentityData(in.Identity),
+		Diagnostic:     Diagnostics(in.Diagnostics),
```
ansgarm, line 26: Can we fix `Diagnostic` to be plural for consistency reasons? I think that's what we use everywhere

ansgarm: Has to be fixed in the protocol (I think core has it "wrong" [too](https://github.com/hashicorp/terraform/blob/2c721982099c560a353fd3d9ddd87c3f60e93f7e/docs/plugin-protocol/tfplugin5.proto#L847))

tfprotov5/tf5server/server.go
```diff
@@ -1255,3 +1247,65 @@ func (s *server) ValidateListResourceConfig(ctx context.Context, protoReq *tfplu
...
+	if err != nil {
+		logging.ProtocolError(ctx, "Error from downstream", map[string]interface{}{logging.KeyError: err})
+		return err
+	}
+
+	for ev := range resp.Results {
+		select {
+		case <-ctx.Done():
+			logging.ProtocolTrace(ctx, "Context done")
+			return nil
```
ansgarm, line 1287: Is this the cancellation handling? If yes, should we make the Trace more clear?

ansgarm (dismissed):
I like! I only commented my (mostly nits) on v5 – v6 should be changed in a similar way, ofc

examples/terraform-provider-primes/README.md
austinvalle, line 1: Is this example just for your own testing / it's temporary?

If we want to provide examples of full provider server implementations I think it'd need to be more complete, which is tough to maintain (see `terraform-plugin-testing` 😆 )

I do like all the information here tho 👀

austinvalle (dismissed):
LGTM, just had a couple thoughts

examples/terraform-provider-primes/provider.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var _ tfprotov5.ProviderServerWithListResource = PrimeNumberProvider{} //nolint:staticcheck -- yes, this is temporary interface
```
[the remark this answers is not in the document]
bbasata: it seems i need to remove the commentary after the ignore

examples/terraform-provider-primes/provider.go
bbasata: >  Is this the case where the caller effectively cancels?

Yes! I think I wrote a code comment (in Framework?) that documents this and I will port it over here.

tfprotov6/list_resource.go
```diff
@@ -5,15 +5,71 @@ package tfprotov6
...
+// to the provider's ListResource implementation. The provider should set a
+// Results iterator function that pushes zero or more results of type
+// ListResourceResult.
+//
+// For convenience, a provider implementation may choose to convert a slice of
+// results into an iterator using [slices.Values].
+//
+// [slices.Values]: https://pkg.go.dev/slices#Values
+type ListResourceServerStream struct {
+	Results iter.Seq[ListResourceResult]
```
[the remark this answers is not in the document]
bbasata: 👍 so many comments 😄

austinvalle (approved):

PR #531, opened 2025-06-27
[DRAFT] Support for PSS

radeksimko:
----

⚠️  Not to be merged at this point ⚠️ 

---

Only for prototyping and discussion for now.

### DIFF
--- a/tfprotov6/tf6server/server.go
+++ b/tfprotov6/tf6server/server.go
@@ -1504,6 +1504,191 @@ func (s *server) InvokeAction(protoReq *tfplugin6.InvokeAction_Request, protoStr
 	return nil
 }
 
+func (s *server) ValidateStateStoreConfig(ctx context.Context, protoReq *tfplugin6.ValidateStateStore_Request) (*tfplugin6.ValidateStateStore_Response, error) {
+	rpc := "ValidateStateStoreConfig"
+	ctx = s.loggingContext(ctx)
+	ctx = logging.RpcContext(ctx, rpc)
+	ctx = logging.ResourceContext(ctx, protoReq.TypeName)
+	ctx = s.stoppableContext(ctx)
+	logging.ProtocolTrace(ctx, "Received request")
+	defer logging.ProtocolTrace(ctx, "Served request")
+
+	req := fromproto.ValidateStateStoreRequest(protoReq)
+
+	logging.ProtocolData(ctx, s.protocolDataDir, rpc, "Request", "Config", req.Config)
+
+	ctx = tf6serverlogging.DownstreamRequest(ctx)
+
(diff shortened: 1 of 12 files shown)


tfprotov6/tf6server/server.go
```diff
@@ -1504,6 +1505,310 @@ func (s *server) InvokeAction(protoReq *tfplugin6.InvokeAction_Request, protoStr
...
+				Bytes:       chunk.Bytes,
+				TotalLength: chunk.TotalLength,
+				Range: tfprotov6.StateByteRange{
+					Start: chunk.Range.Start,
+					End:   chunk.Range.End,
+				},
+			})
+			if !ok {
+				return
+			}
```
SarahFrench, line 1704: I've been looking at this part and how to handle the errors encountered here. I think we can use the iter.Seq2 type to just return the errors to the calling code that uses the iterator, and the error can be handled there.

This would mean updating the `WriteStateBytesStream` struct:

```diff
type WriteStateBytesStream struct {
-	Chunks iter.Seq[WriteStateByteChunk]
+	Chunks iter.Seq2[WriteStateByteChunk, error]
}
```

This allows the iterator to return two values when it's used with `range`, and the calling code will need to handle when the err isn't nil.

```go
for chunk, err := range stream.Chunks {
    if err != nil {
        // handle error.
        // The io.EOF is never returned from the iterator; when io.EOF happens it stops the for loop.
    }
    
    // Use chunk value...
}

```

Changes to the iterator would look like this:

```diff
			if err != nil {
-				// attempt to send the error back to client
-				msgErr := srv.SendMsg(&tfplugin6.WriteStateBytes_Response{
-					Diagnostics: toproto.Diagnostics([]*tfprotov6.Diagnostic{
-						{
-							Severity: tfprotov6.DiagnosticSeverityError,
-							Summary:  "Writing state chunk failed",
-							Detail: fmt.Sprintf("Attempt to write a byte chunk of state %q to %q failed: %s",
-								chunk.StateId, chunk.TypeName, err),
-						},
-					}),
-				})
-				if msgErr != nil {
-					err := status.Error(codes.Unimplemented, "ProviderServer does not implement WriteStateBytes")
-					logging.ProtocolError(ctx, err.Error())
-					return
-				}
-				return
+				// A non-sentinel error has ocurred
+				logging.ProtocolError(ctx, fmt.Sprintf(
+					"WriteStateBytes experienced an error when receiving state data from Terraform: %s",
+					err),
				)
			}

			ok := yield(tfprotov6.WriteStateByteChunk{
				Bytes:       chunk.Bytes,
				TotalLength: chunk.TotalLength,
				Range: tfprotov6.StateByteRange{
					Start: chunk.Range.Start,
					End:   chunk.Range.End,
				},
+			}, err)
-			})
			if !ok {
				return
			}
```

PR #532, opened 2025-07-02
Update `ListResource` to include `limit`

stephybun:
## Description

Updates the proto definitions for 5 and 6 to include `limit` in `ListResource`

## Rollback Plan

None


austinvalle (approved):
LGTM - Might be already on your radar, but to fully complete this you'll also want to add limit to the exported request type and the mapping logic:
- `tfprotov{5/6}/internal/fromproto/list_resource.go`
- `tfprotov{5/6}/list_resource.go`

austinvalle (approved):
🚀

PR #533, opened 2025-07-03
Record changelogs for List RPCs

bbasata:
## Description

This pull request adds changelog entries for an upcoming pre-release.

## Rollback Plan

None.

### DIFF
--- a/tfprotov5/internal/fromproto/list_resource.go
+++ b/tfprotov5/internal/fromproto/list_resource.go
@@ -17,6 +17,7 @@ func ListResourceRequest(in *tfplugin5.ListResource_Request) *tfprotov5.ListReso
 		TypeName:        in.TypeName,
 		Config:          DynamicValue(in.Config),
 		IncludeResource: in.IncludeResourceObject,
+		Limit:           in.Limit,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


.changes/unreleased/NOTES-20250703-163653.yaml
```diff
@@ -0,0 +1,5 @@
+kind: NOTES
+body: These definitions are considered experimental and may change up until general availability.
```
austinvalle, line 2: Just a drive-by comment 🚗 , since we're also queuing up other protocol changes like https://github.com/hashicorp/terraform-plugin-go/pull/534, we should probably be more specific on what definitions are considered experimental.

```suggestion
body: The list resource protocol definitions are considered experimental and may change up until general availability.
```

.changes/unreleased/NOTES-20250703-163637.yaml
```diff
@@ -0,0 +1,5 @@
+kind: NOTES
+body: A `ProviderServerWithListResource` can be used with the `terraform query `subcommand in a future Terraform pre-release to search unmanaged infrastructure.
```
bbasata, line 2:
```suggestion
body: A `ProviderServerWithListResource` can be used with the `terraform query` subcommand in Terraform 1.13.0-alpha20250708 and later to search unmanaged infrastructure.
```

PR #534, opened 2025-07-07
tfprotov5+tfprotov6: Add protocol support for planning/invoking actions

austinvalle:
## Related Issue

N/A

## Description

This PR introduces the protocol definition and initial exported types for defining, planning, and invoking actions via Terraform. In addition to new schema/metadata types for actions, the new RPCs introduced are `PlanAction` and `InvokeAction`.

A temporary `ProviderServerWithActions` interface has been introduced and will be removed in future versions, which allows downstream provider servers to not implement these new RPCs for a couple versions.

------

I didn't add any changelogs to this PR since there aren't any ways to use this protocol ATM, we can come back later and add them 👍🏻 

## Rollback Plan

No


tfprotov5/action.go
```diff
@@ -0,0 +1,254 @@
...
+	// safely accessing the configuration.
+	Config *DynamicValue
+
+	// ClientCapabilities defines optionally supported protocol features for the
+	// PlanAction RPC, such as forward-compatible Terraform behavior changes.
+	ClientCapabilities *PlanActionClientCapabilities
+}
+
+// ProposedLinkedResource represents linked resource data before PlanAction is called.
+type ProposedLinkedResource struct {
```
austinvalle, line 64: `ProposedLinkedResource`, `PlannedLinkedResource`, `InvokeLinkedResource`, and `NewLinkedResource` are the only deviations from the protocol, so I wasn't really sure what a good name for those structs would be.

These are all of the embedded message definitions for linked resources, so we could alternatively just embed the struct definition in each corresponding parent to exactly match the protocol (in other places we have always broken them out, but here there is so much conflicting naming, not sure which is more confusing 🤔 )

rainkwan: 🤔

tfprotov5/internal/toproto/action.go
```diff
@@ -0,0 +1,94 @@
...
+					Diagnostics:     Diagnostics(event.Diagnostics),
+				},
+			},
+		}
+	}
+
+	// It is not currently possible to create tfprotov5.InvokeActionEventType
+	// implementations outside the tfprotov5 package. If this panic was reached,
+	// it implies that a new event type was introduced and needs to be implemented
+	// as a new case above.
```
SBGoods, line 78: Nice comment 👍🏾

tfprotov5/action.go
SBGoods: I think the way that you have it broken up makes the most sense. I'm a bit worried that if we embed a `LinkedResource` struct in each of the RPC request structs that it will make the comments unreadable because of the indentation and it will bloat the parent struct.

tfprotov5/action_schema.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// LinkedResourceSchema represents information about the schema of a linked resource, which is used by an action schema to describe to Terraform the
+// resource types that an action is allowed to change the state of. Linked resources are currently restricted to be defined in the same provider
+// as the action is defined.
+//
+// LinkedResourceSchema does not contain the entire schema definition of the linked resource, which must be obtained by the provider in order to
+// decode the linked resource plan/state/identity protocol data during PlanAction and InvokeAction.
+type LinkedResourceSchema struct {
+	// TypeName is the name of the managed resource which can have it's resource state changed by the action.
```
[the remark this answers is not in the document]
austinvalle: Good call, added in ded2b7eeb44b93412ee1fb3e47d4344e6245abf6 ⚡

PR #538, opened 2025-07-16
Update `ValidateListResourceConfig` to validate `include_resource` and `limit`

stephybun:
## Description

Updates the proto definitions for 5 and 6 to validate `include_resource` and `limit` 

## Rollback Plan

None

### DIFF
--- a/tfprotov6/list_resource.go
+++ b/tfprotov6/list_resource.go
@@ -109,6 +109,15 @@ type ValidateListResourceConfigRequest struct {
 	// knowing the value at request time. Any attributes not directly set in
 	// the configuration will be null.
 	Config *DynamicValue
+
+	// IncludeResourceObject is the value of the include_resource
+	// argument in the list block. This is a DynamicValue so that it can
+	// contain unknown values.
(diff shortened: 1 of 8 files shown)


tfprotov6/list_resource.go
```diff
@@ -109,6 +109,15 @@ type ValidateListResourceConfigRequest struct {
 	// knowing the value at request time. Any attributes not directly set in
 	// the configuration will be null.
 	Config *DynamicValue
+
+	// IncludeResourceObject is the value of the include_resource
+	// argument in the list block. This is a DynamicValue so that it can
+	// contain unknown values.
```
bbasata, line 115: This was my first look at the updated proto files, so the `DynamicValue`s for primitive-or-unknown in the protocol surprised me. It feels like a rare example/precedent – a `DynamicValue` typically represents a value with a provider-defined schema type.

I do recall the discussion of allowing expressions in config, so this may indeed be necessary. My one thought here is to consider how to simplify access to these fields in Framework – and possibly plugin-go too – for the provider developer.

bbasata (approved):
:shipit:

PR #554, opened 2025-09-03
Update action proto v5/v6 definitions to match TF core

austinvalle:
## Related Issue

N/A

## Description

This PR brings our proto file up-to-date with TF core, notably just removing unused action schema types and adding linked resource config data to the validate RPC (currently unreachable with unlinked schema types).

There are also some formal reservation of field numbers that will be used by future features, as I didn't know `reserved` was an actual keyword when I wrote the original comment 😆 

## Rollback Plan

No


go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/hashicorp/terraform-plugin-go
 
-go 1.24
+go 1.23.0
```
austinvalle, line 3: Looks like the Go module was inadvertently bumped by `go-plugin@1.7.0`, I'm reverting that change for the moment, so we can come back and bump the version along with the CI -> https://github.com/hashicorp/terraform-providers-devex-internal/issues/189

SBGoods (approved):

PR #563, opened 2025-09-18
tfprotov6: Add protocol support for pluggable state storage

radeksimko:
> _**Maintainer note from @austinvalle 👋🏻**_ 
>
> After discussion offline, this PR is in a good place to be merged now, and all review feedback are small naming adjustments / unit tests / etc, which I will be opening a follow-up PR to reflect. 

This PR introduces the protocol (v6 only) for providers to support Pluggable [State Storage](https://developer.hashicorp.com/terraform/language/state/backends) (PSS), which will be represented as the `state_store` type in configurations. The corresponding changes for PSS support are still on-going across a couple repositories so it's possible these types/names/rpc methods could change prior to a stable release of this Go module.

### DIFF
--- a/tfprotov6/internal/fromproto/state_store.go
+++ b/tfprotov6/internal/fromproto/state_store.go
@@ -85,5 +85,6 @@ func UnlockStateRequest(in *tfplugin6.UnlockState_Request) *tfprotov6.UnlockStat
 	return &tfprotov6.UnlockStateRequest{
 		TypeName: in.TypeName,
 		StateId:  in.StateId,
+		LockId:   in.LockId,
 	}
 }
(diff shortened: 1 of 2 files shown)


SarahFrench:
@radeksimko FYI if you rebase this branch on main it'll pull in changes to how we use protoc, so you can use `make protobuf` in this repo and in hashicorp/terraform when working on both sides of the protocol changes

austinvalle (approved):
LGTM 🚀, I updated the PR description + title, but all review feedback I have are minimal naming/tests/logging related, which I will handle in a follow-up PR.

ansgarm (approved):

PR #565, opened 2025-09-19
Sarah's updates to f-pluggable-state-store : Make naming consistent by avoiding 'Msg'

SarahFrench:
## Related Issue

Fixes # 

## Description

In plain English, describe your approach to addressing the issue linked above. For example, if you made a particular design decision, let us know why you chose this path instead of another solution.

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.

### DIFF
--- a/tfprotov6/state_store.go
+++ b/tfprotov6/state_store.go
@@ -53,7 +53,7 @@ type ReadStateBytesRequest struct {
 }
 
 type ReadStateBytesStream struct {
-	Chunks iter.Seq[ReadStateByteChunk]
+	Messages iter.Seq[ReadStateByteStreamMsg]
 }
 
 type WriteStateBytesStream struct {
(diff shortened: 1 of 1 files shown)


tfprotov6/state_store.go
```diff
@@ -57,24 +57,24 @@ type ReadStateBytesStream struct {
...
 //     receiving data from Terraform core.
-type WriteStateBytesStreamMsg struct {
-	Chunk WriteStateByteChunk
-	Err   error
+//
+// If a gRPC error is set, then the chunk should be empty.
+type WriteStateBytesChunk struct {
+	StateByteChunk
+	Err error
 }
```
SarahFrench, line 72: This now matches `ReadStateByteChunk` below

radeksimko:
I do admit the `Msg`/message part of the name did make me pause and think but I could not come up with any substantially better name or solution to object it 😅 

I suppose consistency is always a valid argument.

